PR #59, opened 2019-09-07
MINOR: move ConfigMap to controller namespace

yann-soubeyrand:
ConfigMap should be located alongside the Deployment, it's common practice and eases uninstallation: you only need to delete the namespace.

### DIFF
--- a/deploy/haproxy-ingress.yaml
+++ b/deploy/haproxy-ingress.yaml
@@ -80,8 +80,8 @@ subjects:
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: haproxy-configmap
-  namespace: default
+  name: haproxy
+  namespace: haproxy-controller
 data:
 
 ---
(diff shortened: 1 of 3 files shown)


yann-soubeyrand:
IMHO, the default should be to look for the ConfigMap in the same namespace the controller runs in. This is not something which could be easily expressed using flags default value. This should happen further in the code. Having this default would also allow to remove the --configmap option in YAML files. Ultimately, I thing the YAML files should have an empty args section and rely only on default values for configuration options. This allows people to easily customize deployments by adding options to this args section without having to worry about merging existing and new options.

bedis:
I think that the default should be looking for the config map in the same namespace too. and the ingress controller should not run in the default namespace by default too, because your secrets have to be in that same namespace.

Helm makes it easy to deploy an app in a container and using variables in your template to make it consistent is trivial to do as well.

oktalz:
hi @yann-soubeyrand 
so, this will have to wait a bit (few days) until i remove restriction of even having configmap as mandatory (even an empty one). There are few ways how to do this and I haven't decided yet how exactly will i do that.

After that change, i think we can merge this PR.

Mo3m3n:
Hi,
We were able to merge this one after making configMap optional https://github.com/haproxytech/kubernetes-ingress/commit/3e040ec00bd47c402f7e2a1a6d739ff3d429a83c
The controller will still need a full path namespace/name for the configmap though 
So I needed to rebase your change and made it clear in the log message that configMap is moved to controller namespace in the deploy file (but not as an implementation of the controller logic)

PR #60, opened 2019-09-07
MINOR: do not set default-ssl-certificate by default

yann-soubeyrand:
Default SSL certificate should not be enabled by default: it enables SSL redirect which gives errors when the secret is not there (very common setup, since a lot of people make use of cert-manager and don't have default certificate).


oktalz:
Hi @yann-soubeyrand 

Thanks for the PR.
It makes sense to remove that from example.

PR #62, opened 2019-09-09
MINOR: add kustomization.yaml alongside deployment manifests

yann-soubeyrand:
This will allow customizing deployment manifests using kustomize.

### DIFF
--- a/deploy/kustomization.yaml
+++ b/deploy/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - haproxy-ingress.yaml


bedis:
Hi @yann-soubeyrand,

Sorry for my ignorance, but could you please let us know what is kustomize and how it compares to helm from your point of view?
Is your patch "good" enough for kustomize or is it fully supported / integrated?

yann-soubeyrand:
Hi @bedis,

Kustomize is a tool to customize Kubernetes manifests in a purely declarative fashion (as opposed to Helm templating). They are not direct competitors in my opinion, they just have different approaches.

Simply adding a kustomization.yaml file allows for the manifests to be referenced remotely and customized by kustomize users. An example can be found here https://github.com/argoproj/argoproj-deployments/blob/master/argocd/kustomization.yaml: you can see that this kustomization.yaml references a remote repository (https://github.com/argoproj/argo-cd/tree/master/manifests/ha/cluster-install) as its base, and then adds resources to customize the deployment.

In addition to being able to add new resources, you can patch resources (for example adding entries to the config map). For simple customizations, it can be simpler to use than Helm.

Furthermore, kustomize is now part of kubectl since version 1.14: you can directly run `kubectl apply -k .` in a directory which contains kustomizations.yaml files.

If you want further examples of kustomize utilization, you can have a look at its use for deploying NGINX ingress controller (https://github.com/kubernetes/ingress-nginx/tree/master/deploy/aws/l4).

bedis:
Hey, sorry for long time without answering here.
I think something that would help a lot is a quick documentation on how to use this kustomization and one or two simple example to help people using it.

jcpetruzza:
Having a `kustomization.yaml` file that just imports the manifests mean that other people now *can* use `kustomize` to install the ingress using the provided manifests plus some local modification. E.g., I could then define the following `kustomization.yaml` file locally :

```
$ cat > ingress-controller/kustomization.yaml <<EOF
resources:
- https://raw.githubusercontent.com/haproxytech/kubernetes-ingress/v1.2.2/deploy

namespace: ingress-controller
```

If you then run `kustomize build ingress-controller` you will get the same manifests defined in `deploy/haproxy-ingress.yaml` but with the `namespace` changed to `ingress-controller`. You can then deploy this "customized" version with `kubectl apply -k ingress-controller`.

Having a `kustomization.yaml` in the `deploy` directory is a requirement of `kustomize`, in order to be able to import these resources over the network. But it is an opt-in thing, having it there shouldn't affect people not using `kustomize`.

PR #70, opened 2019-09-18
BUG/MEDIUM: Reset server address key to random string when…

### LINKED ISSUE #69: Switching back to an old Deployment is not registered by controller
## Problem

The controller does not add servers if those servers were added and removed previously when the pod selector was changed (such as to change the 'version' from 1.0 to 1.1, then back 1.0) on the service's Service.

## Environment

* Kubernetes v1.14.2
* Minikube v1.1.0
* haproxytech/kubernetes-ingress (469f92704ba5395ffbd38c10ac32538c23823a14e2ca474abca281ff8b74bae7)

## How to reproduce

1. Add 2 Deployments that are identical except that they have different labels:

*Blue Deployment*:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    run: app
  name: app-blue
spec:
  replicas: 1
  selector:
    matchLabels:
      run: app
      scope: blue
  template:
    metadata:
      labels:
        run: app
        scope: blue
    spec:
      containers:
      - name: app
        image: quay.io/nickmramirez/webapp
        ports:
        - containerPort: 80
```

*Green Deployment*:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    run: app
  name: app-green
spec:
  replicas: 1
  selector:
    matchLabels:
      run: app
      scope: green
  template:
    metadata:
      labels:
        run: app
        scope: green
    spec:
      containers:
      - name: app
        image: quay.io/nickmramirez/webapp
        ports:
        - containerPort: 80
```

2. Add a service that points to *Blue Deployment*:

```
---
apiVersion: v1
kind: Service
metadata:
  labels:
    run: app
  name: app
spec:
  selector:
    run: app
    scope: blue
  ports:
  - name: port-1
    port: 80
    protocol: TCP
    targetPort: 80
```

3. Add an Ingress:

```
---
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: app-ingress
  namespace: default
spec:
  rules:
  - http:
      paths:
      - path: /
        backend:
          serviceName: app
          servicePort: 80
```

4. At this point, the *blue* servers are added to the HAProxy Stats page. Next, change the `Service` so that its *scope* selector is *green*. HAProxy will remove the blue server and add the green server. So far, so good.

5. Change the *scope* selector on the `Service` back to *blue*. This time, the *green* servers are removed, but the *blue* servers are not added.

NickMRamirez:
… modified and disabled.

### DIFF
--- a/controller-events.go
+++ b/controller-events.go
@@ -248,7 +248,7 @@ func (c *HAProxyController) setModifiedStatusEndpoints(oldObj, newObj *Endpoints
 			adrOld.IP = "127.0.0.1"
 			adrOld.Disabled = true
 			adrOld.Status = MODIFIED
-			(*newObj.Addresses)[oldKey] = adrOld
+			(*newObj.Addresses)[fmt.Sprintf("SRV_%s", RandomString(5))] = adrOld
 		} else {
 			//try to find one that is added so we can switch them
 			replaced := false


jaytaylor:
What does this fix, exactly?

NickMRamirez:
@jaytaylor This fixed https://github.com/haproxytech/kubernetes-ingress/issues/69

In which switching back and forth between label selectors on the Service that's targeted by the Ingress results in the server not being added to the backend, but only once you've switched away from it, and then back to it.

1. Set service label to 1.0 pods
2. Set service label to 1.1 pods
3. Set service label to 1.0 pods <- error! Pods not added to backend

PR #117, opened 2019-11-16
enable https frontend when certs are in use

### LINKED ISSUE #116: [dev branch] frontend https listeners not working with existing tls secrets
In the `dev` branch, the controller never enables the https frontend listeners for existing ingresses with TLS secrets:

```
	if len(certsUsed) > 0 {
		usingHTTPS = true
		c.enableCerts() // HACK: force enable certs on every iteration
	}
```

Also, `handleDefaultCertificate` may change `c.UseHTTPS` but because enableCerts is evaluated **before** `handleDefaultCertificate` then it may never enable https listeners for default cert:

Snippet for reference:
```
	if c.UseHTTPS.Status != EMPTY {
		c.enableCerts()
		c.UseHTTPS.Status = EMPTY
	}

	usingHTTPS := false
	reload = c.handleDefaultCertificate(certsUsed)
	needsReload = needsReload || reload
	if len(certsUsed) > 0 {
		usingHTTPS = true
	}
```

/cc @Mo3m3n

bithavoc:
Fixes #116 in the `dev` branch by using a single source of truth when the certs are used for the first time.
Also moves the evaluation of `len(certsUsed) > 0` after `c.handleDefaultCertificate` which has a chance of changing `certsUsed`.


bithavoc:
@Mo3m3n ☝️

Mo3m3n:
Thanks @bithavoc for trying to fix this. However this solves part of the issue because it enables https but does not reload haproxy when new certificates are configured (via k8s secrets) and I was able to confirm it by building an image from your branch and doing the test (adding new secrets after startup).

bithavoc:
Its cool, feel free to merge or close as needed.

PR #129, opened 2019-12-20
make more reliable and scalable ingress 

mohsenmottaghi:
I think ingress container should be deployed as `DaemonSet` to be work distributed and scalable, if you use `Deployment` you should set replica manually and add some attribute for pod anti-affinity to distribute all replicas on all workers but I think `DaemonSet` is a better option for automatic scaling. and `DaemonSet` let us to use `hostPort` to expose HTTP over port 80 and HTTPS over port 443 and we aren`t limit to use port number from 30000 range.

### DIFF
--- a/deploy/haproxy-ingress.yaml
+++ b/deploy/haproxy-ingress.yaml
@@ -127,7 +127,7 @@ spec:
 
 ---
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
 metadata:
   labels:
     run: haproxy-ingress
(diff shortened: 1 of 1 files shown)


oktalz:
hi @mohsenmottaghi 

thx for the PR, maybe its better to create a new file for that kind of setup.
deploy/haproxy-ingress-deamonset.yaml

There are many ways how controller can be created/managed
and it would be better to have more examples so everybody can choose what best suits them.

oktalz:
also if you can follow our [CONTRIBUTING guide](https://github.com/haproxy/haproxy/blob/master/CONTRIBUTING), 

In short what I'm referring to is how we are creating commit messages.
as an example this could be something like 
`BUILD: deployed as DaemonSet with hostPort to improve relaiability`

mohsenmottaghi:
thanks for your suggestions. I close this pull request and create a new pull request in https://github.com/haproxytech/kubernetes-ingress/pull/134 with your suggestions . thank you

PR #132, opened 2020-01-08
BUG/MAJOR: Force writing a newline in PEM certificate bundle

interone-ms:
In some cases there is no newline written between the key and the certificate
bundle. This results in the line being mangled and a parse error upon initialization:
-----END RSA PRIVATE KEY----------BEGIN CERTIFICATE-----

This patch always writes a newline to ensure the key and certificate bundle can be correctly loaded.

### DIFF
--- a/controller-https.go
+++ b/controller-https.go
@@ -44,7 +44,7 @@ func (c *HAProxyController) cleanCertDir(usedCerts map[string]struct{}) error {
 	return nil
 }
 
-func (c *HAProxyController) writeCert(filename string, key []byte, crt []byte) error {
+func (c *HAProxyController) writeCert(filename string, key, crt []byte) error {
 	var f *os.File
 	var err error
 	if f, err = os.Create(filename); err != nil {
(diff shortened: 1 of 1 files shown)


oktalz:
Hi @interone-ms 
Thx for the PR !

Yes, forgetting to put `\n` on end of file/data is something that can happen,
but I wouldn't call this a major bug, it's more of a minor annoyance.

In general it looks good but I would rather that 
we do a check before writing `\n` to see if this is needed or not.

It can easily be checked if last byte in `key` is `\n` or not.
Not having that check would put extra line in file and while this is not an major issue, we should be consistent here.

interone-ms:
I'll look into it, right now I'm busy to track down a bug where the controller adds a backend definition for all workloads in a namespace instead of only the workload I specify in Rancher. I'll update you tomorrow...

interone-ms:
So I did a couple of tests, apparently at least Rancher always trims the SSL certificate before storing the secret object, while kubectl does not do any manipulation.

Now the patch checks if the last byte is a \n and writes it extra, if not.

oktalz:
I would say that omitting \n on any file is a bad practice, 
but luckily we have easy solution for this that you provided.

major or medium is more of a info for all, not just Rancher users,
but i will not complicate this to much :)

merging, thanks again for the changes and for this PR.

PR #143, opened 2020-01-28
MINOR: Added annotation "server-ssl", sets downstream server "ssl verify none"

andreasdrougge:
Added annotation "server-ssl", this sets downstream server "ssl verify none" if enable, default is disable.

### DIFF
--- a/annotations.go
+++ b/annotations.go
@@ -144,4 +144,5 @@ var defaultAnnotationValues = MapStringW{
 	"timeout-http-keep-alive":   &StringW{Value: "1m"},
 	"whitelist":                 &StringW{Value: ""},
 	"whitelist-with-rate-limit": &StringW{Value: "false"},
+	"server-ssl":                &StringW{Value: "false"},
 }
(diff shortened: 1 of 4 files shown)


Mo3m3n:
Hey @andreasdrougge 
Thanks for contributing!, can you just add the following:
- Update [doc](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/README.md) to add the server-ssl annotation. By the way the scope of this one would be Config-map and Service.
- Change the commit message to meet the format we use, so in this case you add prefix "MINOR:"
And we will merge the PR as soon as the changes are made.

annotations.go
```diff
@@ -144,4 +144,5 @@ var defaultAnnotationValues = MapStringW{
 	"timeout-http-keep-alive":   &StringW{Value: "1m"},
 	"whitelist":                 &StringW{Value: ""},
 	"whitelist-with-rate-limit": &StringW{Value: "false"},
+	"server-ssl":                &StringW{Value: "false"},
```
oktalz: except `ingress.class`

all other annotations are in alphabetical order, please move server-ssl to appropriate position

andreasdrougge: Fixed now

Mo3m3n:
- You have changed the PR message but I was rather talking about the "commit" message, the one you see in git commits
- No need to have three different commits for a minor feature. We were expecting your commit locally, rebase and resend the pull request.
- The formatting of the docs is not quite correct (extra spaces, indentation, etc).
I have applied these changes manually and merged your commit: https://github.com/haproxytech/kubernetes-ingress/commit/6a71ded212896b237f8990b191ac58223bd15d32

PR #145, opened 2020-01-30
MINOR: Added cookie settings, support for sticky session

andreasdrougge:
Added support for cookie, sticky session

### DIFF
--- a/documentation/README.md
+++ b/documentation/README.md
@@ -12,6 +12,17 @@ Options for starting controller can be found in [controller.md](controller.md)
 
 | Annotation | Type | Default | Dependencies | Config map | Ingress | Service |
 | - |:-:|:-:|:-:|:-:|:-:|:-:|
+| [cookie](#cookie) | string | "" |  |:large_blue_circle:|:large_blue_circle:|:large_blue_circle:|
+| [cookie-domain](#cookie) | string | "" | [cookie](#cookie) |:large_blue_circle:|:large_blue_circle:|:large_blue_circle:|
+| [cookie-dynamic](#cookie) | ["true", "false"] | "false" | [cookie](#cookie) |:large_blue_circle:|:large_blue_circle:|:large_blue_circle:|
+| [cookie-indirect](#cookie) | ["true", "false"] | "true" | [cookie](#cookie) |:large_blue_circle:|:large_blue_circle:|:large_blue_circle:|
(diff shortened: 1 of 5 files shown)


documentation/README.md
```diff
@@ -25,7 +36,7 @@ Options for starting controller can be found in [controller.md](controller.md)
 | [rate-limit-expire](#rate-limit) | string | "30m" | [rate-limit](#rate-limit) |:large_blue_circle:|:white_circle:|:white_circle:|
 | [rate-limit-interval](#rate-limit) | string | "10s" | [rate-limit](#rate-limit) |:large_blue_circle:|:white_circle:|:white_circle:|
 | [rate-limit-size](#rate-limit) | string | "100k" | [rate-limit](#rate-limit) |:large_blue_circle:|:white_circle:|:white_circle:|
-| [server-ssl](#server-ssl) | ["true", "false"] | "false" |  |:large_blue_circle:|:white_circle:|:large_blue_circle:|
+| [server-ssl](#server-ssl) | ["true", "false"] | "false" |  |:large_blue_circle:|:large_blue_circle:|:large_blue_circle:|
```
oktalz: if this was wrongly documented, i would prefer that you do a separate commit (BUG/MINOR: server-ssl)
since this has nothing to do with cookies and if it slipped, remove it.

andreasdrougge: True, I'll fix

Mo3m3n:
@andreasdrougge 
I have made some changes to your work:
- Renaming cookie to cookie-persistence, to be more precise about the annotation role.
- Exposing all the cookies annotation in the doc may be confusing (having the majority of annotations talking about cookies while using cookie persistence is a special usecase), so we though that it is better to describe only the main annotation in the doc and point the user to code for more annotations.
- Also in the code, it is better to separate annotations used to fine tune cookies.
Let me know if you're ok with that or not.
You can find here the commit in question: https://github.com/haproxytech/kubernetes-ingress/commit/14694591e23425c496e3c463b3c305cd50b06e2d

andreasdrougge:
>   Let me know if you're ok with that or not.
>   You can find here the commit in question: [1469459](https://github.com/haproxytech/kubernetes-ingress/commit/14694591e23425c496e3c463b3c305cd50b06e2d)
Looks good! Go ahead and use this.

Mo3m3n:
Thanks PR applied here https://github.com/haproxytech/kubernetes-ingress/commit/6ed932d010d0abcb59bd17856189a0b77a84aba4

Issue #124, opened 2019-12-02
Backend services unaware of request scheme


easkay-castlight:
Hi folks,

I have a webapp running in kubernetes, with HAProxy as the ingress controller. The webapp uses the incoming request (including headers) to determine what scheme to use when generating URLs. I've noticed that when accessing the app via the ingress controller, some links are HTTP instead of HTTPS, meaning that the app is unaware of what scheme was used for the original request.

In the past, I have solved this issue by setting `X-Forwarded-Proto` to `https` to inform the webapp that the original request was HTTPS and therefore the generated links are HTTPS as well.

Is this something that the ingress controller could support? I believe a relatively simple set-header in the frontend would do the trick:
```
http-request set-header X-Forwarded-Proto https if { ssl_fc }
```
(adapted from https://stackoverflow.com/questions/51928504/x-forwarded-proto-https-in-frontend-or-backend-haproxy).

I'd be happy to raise a PR to add the functionality if you'd like.

Thanks for your consideration.

oktalz:
hi @easkay-castlight 
this seem like a good option to have

PR is definitely welcomed here :)

easkay-castlight:
Cool, thanks @oktalz. I'll try and have a look soon.

easkay:
@oktalz I have a couple of ideas about where/how to implement this, but I had a couple of questions:

* Should this behaviour have an option to control it? E.g. configmap/annotations/CLI args
* If not, where would be the best place to add the HTTPRequestsRule? I haven't yet fully studied the code and as yet don't know what happens at what point, so some advice would be greatly appreciated

If it's better to discuss these things in a PR, then I'm happy to put a draft in.

easkay-castlight:
This was fixed in #146.

PR #146, opened 2020-01-30
MINOR: Send X-Forwarded-Proto to HTTP backends for TLS clients.

easkay:
Hi folks,

Thanks for taking the time to look at this, my apologies if I've misunderstood something or messed something up - this is my first PR here.

This PR addresses #124 by setting the `X-Forwarded-Proto` header with a value of `https` if the incoming request was received over TLS.

I wasn't immediately sure about how to ensure this configuration stayed constant, so I tried to follow what looked like existing patterns. If desired, I'm sure this behaviour could be configurable as opposed to on by default with no option to disable. I'd be happy to discuss it.

Most of all, I'd be very grateful for feedback on how I can improve the code here, and/or whether there's a better approach. Some pointers for where to add tests would be appreciated too.

Thanks in advance for your time.

### DIFF
--- a/configuration.go
+++ b/configuration.go
@@ -26,8 +26,6 @@ const (
 	RATE_LIMIT = "rate-limit"
 	//nolint
 	HTTP_REDIRECT = "http-redirect"
-	//nolint
-	X_FORWARDED_PROTO = "x-forwarded-proto"
 )
 
 //Configuration represents k8s state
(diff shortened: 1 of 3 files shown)


Mo3m3n:
Hi @easkay 
Thank your for the time spent on this.
I have tested it and it is working as expected.
Few comments for future PR maybe  :) :
- Instead of `id := int64(0)` and  `&id`  we have the helper function `ptrInt64(0)`
- Commit messages are prefixed with the patch type (https://github.com/haproxy/haproxy/blob/master/CONTRIBUTING#L632) in this case It is just "MINOR".
I have done these changes and applied the PR here: https://github.com/haproxytech/kubernetes-ingress/commit/2574d118e01a3772adebb1dfeaea7fd3773eec4b

easkay:
Hi @Mo3m3n, thank you very much for the time you've taken to look this over. I wasn't sure what the appropriate patch level was, so thank you for the correction, it's much appreciated.

I did in fact notice the `ptrInt64()` function and I thought I'd applied it, but perhaps I didn't push my changes, apologies.

Thanks for the merge!

PR #149, opened 2020-02-03
BUG/MINOR: server-ssl, fixed documentation

andreasdrougge:
Fixed documentation error. server-ssl work for ingress too.


Mo3m3n:
The old version is the correct one, since server-ssl is configured in the backend (service) it cannot be applied by ingress (to a set of hosts), even if you set it at the ingress level it will apply to the services wherever they are used (even in other ingresses not using the annotation)
Sorry for the confusion, and the doc is not clear and correct about this, we will address this in the near future.

Issue #165, opened 2020-03-06
Empty ingress.class is never ignored


ohardy:
Hi,

It looks like the issue #31 is back for me.

I have two ingress controllers, one for internal domains, one for public domains.

So I configure public one with ingress.class = 'public' for example.

If I add ingress without ingress.class, this ingress controller will not ignore it !

Is it wanted ? Because for me it must not happen.

Thanks in advance

Mo3m3n:
Thanks @ohardy  for another bug report.
It was fixed in related PR and available at `k8s.local/kubernetes-ingress:dev` while we prepare next release

ohardy:
It's fixed in the dev image, thanks !

Mo3m3n:
Released in `haproxytech/kubernetes-ingress:1.4.0`

PR #168, opened 2020-03-12
BUG/MEDIUM: Process only ingress objects which ingress.class strictly matches the controller --ingress.class CLI arg.

Mo3m3n:
In addition to strictly matching ingress.class, controller will dynamically
drop or take into account ingress objects depending on their
ingress.class annotation.

### DIFF
--- a/controller/annotations.go
+++ b/controller/annotations.go
@@ -127,7 +127,6 @@ func SetDefaultAnnotation(annotation, value string) {
 }
 
 var defaultAnnotationValues = MapStringW{
-	"ingress.class":           &StringW{Value: ""},
 	"check":                   &StringW{Value: "true"},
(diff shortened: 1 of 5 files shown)


Mo3m3n:
The strict matching of the ingress.class was reported  in #165 .
Although it was fixed before in https://github.com/haproxytech/kubernetes-ingress/commit/8e20146981ff4e586b38e0cdc9a7cf8f089aea46 a regression was made after that: https://github.com/haproxytech/kubernetes-ingress/commit/e1a53cc3d49281b268d515e26c00a987f4a77feb

This PR will also fix the following:
- Do not track ingress objects with unrelated class
- Dynamically update (delete or apply) ingress object when the ingress.class changes.

PR #169, opened 2020-03-24
MAJOR: Add weights via pod object annotations

dspitzer:
This commit adds support for weights for backend endpoints via pod
annotations. By adding a haproxy.org/weight annotation to a pod you can
specify an explicit weight whenever that pod is being used in an
endpoint as long as the TargetRef was added properly (usually
automatically via services). Additionally, the default weight can be
specified via the new --default-weight parameter.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	clientnative "github.com/haproxytech/client-native"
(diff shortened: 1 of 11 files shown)


dspitzer:
I started a discussion as requestes by Moemen in issue #170

controller/controller.go
```diff
@@ -229,11 +230,45 @@ func (c *HAProxyController) handlePath(namespace *Namespace, ingress *Ingress, r
...
+				}
+			}
+			switch podStatus {
+			case MODIFIED:
+				fallthrough
+			case DELETED:
+				fallthrough
+			case ADDED:
+				needReload = true
+			}
```
oktalz: what do you think about replacing this switch with
```
switch podStatus {
	case MODIFIED, DELETED, ADDED:
		needReload = true
}
```
fallthrough is ok, but I think its more readable without it

controller/controller.go
```diff
@@ -247,6 +282,16 @@ func (c *HAProxyController) handleEndpointIP(namespace *Namespace, ingress *Ingr
 			status = MODIFIED
 		}
 	}
+	if status == EMPTY {
+		switch podStatus {
+		case MODIFIED:
+			fallthrough
```
oktalz: same comment as above

controller/controller.go
dspitzer: Done

controller/controller.go
dspitzer: Done

dspitzer:
Unfortunately this couldn't be finished due to internal developments within the company that lead to no more time allocations and scrapping of the infra involved. Really sorry for this and thanks for all the help.

PR #180, opened 2020-04-25
FEATURE/MEDIUM: Add response-set-header config

### LINKED ISSUE #79: http-response set-header
Would be great to allow admins to set response header both globally and per service / ingress rule.

https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#4.2-http-response%20set-header

easkay:
Fixes #79.

Duplicates the mechanism for `request-set-header` to set response headers, also renamed the `FrontendHTTPRules` map for clarity.

I've tested that the functionality behaves as intended at least in terms of setting headers for the appropriate hostnames given ingress annotations or configmap entries. A pre-built image is available publicly on Dockerhub at [`easkay/haproxy-ingress:response_set_header`](https://hub.docker.com/repository/docker/easkay/kubernetes-ingress).

Thanks in advance for your time, and feedback is much appreciated.

### DIFF
--- a/controller/api.go
+++ b/controller/api.go
@@ -165,9 +165,22 @@ func (c *HAProxyController) frontendHTTPRequestRuleDeleteAll(frontend string) {
 	}
 }
 
+func (c *HAProxyController) frontendHTTPResponseRuleDeleteAll(frontend string) {
+	c.ActiveTransactionHasChanges = true
+	var err error
+	for err == nil {
+		err = c.NativeAPI.Configuration.DeleteHTTPResponseRule(0, "frontend", frontend, c.ActiveTransaction, 0)
+	}
+}
+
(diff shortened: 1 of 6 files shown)


Mo3m3n:
Thank @easkay 
Nice PR which looks good and clean from first look.
We will give it soon a more thorough look and test and eventually merge it.

controller/api.go
```diff
@@ -165,11 +165,24 @@ func (c *HAProxyController) frontendHTTPRequestRuleDeleteAll(frontend string) {
 	}
 }
 
+func (c *HAProxyController) frontendHTTPResponseRuleDeleteAll(frontend string) {
+	c.ActiveTransactionHasChanges = true
+	var err error
+	for err == nil {
+		err = c.NativeAPI.Configuration.DeleteHTTPResponseRule(0, "frontend", frontend, c.ActiveTransaction, 0)
+	}
```
oktalz, line 173: err is always nil then ?

PR #184, opened 2020-04-27
Feature suggestion:  Add ability to inject use_backend rules via annotation

zvickery:
For reasons that I cannot change, I have a need to route traffic with haproxy where the Host header is an IP address.  This is not supported by Kubernetes ingresses because all host names in the ingress specification must be names and not IP addresses.  Basically, for certain ingresses I need the ability to have a rule like `use_backend my-backend if { req.hdr(host),field(1,:) -i 1.2.3.4 } { path_beg /my/path }`.

I have been able to get this working with this PR.  Along with this code running, I have an ingress annotation present which looks like `haproxy.org/ip-routing: 1.2.3.4,5.6.7.8`.  If this functionality is desired in the mainstream ingress controller I'd love to see it merged in some form.

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -186,33 +186,44 @@ func (c *HAProxyController) handleService(namespace *Namespace, ingress *Ingress
 		return backendName, newBackend, reload, nil
 	}
 
-	// Update backendSwitching
-	key := fmt.Sprintf("%s-%s-%s-%s", rule.Host, path.Path, namespace.Name, ingress.Name)
-	useBackendRule := UseBackendRule{
-		Host:      rule.Host,
-		Path:      path.Path,
-		Backend:   backendName,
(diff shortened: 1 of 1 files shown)


zvickery:
Thank you.  I agree about the naming and am not particular about the annotation name.  I look forward to integrating this once the refactoring is complete.

Mo3m3n:
@zvickery 
One year later we implemented this properly via [route-acl](https://github.com/haproxytech/kubernetes-ingress/tree/master/documentation#route-acl) :) 
A service annotation is more appropriate than an Ingress annotation. Because this is a backend/service specific requirement.
Also the suggestion in this PR only deals with a specific case "HostHeader is an IP address", with route-acl you can have any condition that can be described with haproxy acl.

PR #186, opened 2020-05-02
MINOR: Ignore the port part when matching for HTTP redirects

knatsakis:
Similarly to: https://github.com/haproxytech/kubernetes-ingress/commit/aac95d5fef1377b26df9cde6d49efeb0b5a13001


knatsakis:
@Mo3m3n could you please assist me here?

I cannot figure out why the "checking code and commit subject / checking linting and commit subject" check is failing.

Mo3m3n:
Hi @knatsakis 
Thanks for your PR.
I have re-run the checks and they are failing at checkout-code step.
So don't bother with this. Your commit looks even with your first commit message so am merging it and will investigate this later 
Thanks again for the contribution

PR #189, opened 2020-05-11
BUG/MEDIUM: Fix response-set-header behaviour

easkay:
#180 introduced the feature where HTTP response headers can be set. This feature does not work, since the `http-response set-header` rule is combined with an ACL which uses request data, `req.hdr(Host)`. This is invalid HAProxy config and any `http-response set-header` rules get ignored as a result.

This PR moves the `http-response set-header` rule to the relevant backend(s), using a duplicate of the existing backend request modification mechanism. The functionality to handle HTTP response rules in frontends as introduced in #180 has also been removed.

A pre-built Docker image is available at [easkay/kubernetes-ingress:response_set_header](https://hub.docker.com/repository/docker/easkay/kubernetes-ingress).

Thank you for your time, and my sincerest apologies for introducing broken functionality.


easkay:
Screenshot_2020-05-12_09-12-02

End-to-end test having set the `x-foo-header` header to `Bar` and some of the cURL output.

easkay:
Updated to fix conflicts with master.

PR #198, opened 2020-06-04
resolves #197 by setting the default value for sync-period

### LINKED ISSUE #197: SyncPeriod default value not honored
Hi there,

on version 1.4.4 we noticed a problem when running the controller without any cmd line args.
`SyncPeriod` default value does not seem to be working, there is an indeterminate period of time waiting for the monitoring loop to start refreshing.
The default value is defined in the `utils.OSArgs` meta-tags but apparently is not parsed correctly...
I send a PR with some tests and my findings.

inge4pres:
Added an entry in the `defaultAnnotations` map used to configure the controller.
Addresse #197 

It's a checkout of 1.4.4 with the additional code that was tested working


oktalz:
hi @inge4pres 

thx for the PR.
I do have some remarks.

PR can't be merged because it have conflicts with master branch.
Also please see our [CONTRIBUTING guide](https://github.com/haproxy/haproxy/blob/master/CONTRIBUTING).

In short, commit message should have prefix same as all other messages in repo.

Also, having `#197` in message in not desired. If you clone/fork a project that number will automatically make a link to that other workspace and issue 197 might be a different one (or not event exists there).

It better to use comments and comment her like this #197 for this purpose. That way we are having link.

Other remarks i have already put there.

inge4pres:
Thanks for reviewing, I will close this PR and the issue in favour of the upcoming promise of testing the controller config init

PR #214, opened 2020-07-02
remove doubles from rbac role

cruwe:
Apparently, a double declaration of the extensions API-group made it into both the daemonset as well as the deployment example.

I have taken the liberty to suggest the removal of the latter.

In any case, thank you very much for developing the haproxy ingress-controller and making your efforts available to the public. Cheers!

### DIFF
--- a/deploy/haproxy-ingress-daemonset.yaml
+++ b/deploy/haproxy-ingress-daemonset.yaml
@@ -52,14 +52,6 @@ rules:
   - create
   - patch
   - update
-- apiGroups:
-  - extensions
-  resources:
-  - ingresses
-  verbs:
-  - get
-  - list
-  - watch
 
 ---
 kind: ClusterRoleBinding
--- a/deploy/haproxy-ingress.yaml
+++ b/deploy/haproxy-ingress.yaml
@@ -52,14 +52,6 @@ rules:
   - create
   - patch
   - update
-- apiGroups:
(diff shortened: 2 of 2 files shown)


dkorunic:
Make sure to be in sync with actual RBAC policy in Helm chart: https://raw.githubusercontent.com/haproxytech/helm-charts/master/kubernetes-ingress/templates/clusterrole.yaml

cruwe:
I do not quite follow. What I did was to remove the obvious doubles. 

I see the ClusterRole from the helm-chart semantically differs from manifests under `deploy` regarding resources and verbs, but without detailed knowledge about the controller, I do not feel confident changing that and could only do a c&p-job. Perhaps it might be better to entirely remove the manifests if the helm chart is where all maintenance is done?

oktalz:
Hi, @cruwe 

this indeed seems like a duplicate here. 
Our linter showed some issues with commit message
```
Incorrect message format 'remove doubles from rbac role'
Please refer to https://github.com/haproxy/haproxy/blob/master/CONTRIBUTING#L632
```

For commit messages and general style please follow the haproxy project's CONTRIBUTING guide.

cruwe:
Hi ... please excuse my late reaction. Regarding linting, I'm sorry, I could have noted the general idea myself just by looking at the commit log. Thanks for the hint. Cheers!

PR #217, opened 2020-07-10
Making Kubernetes Shared Informer resync period configurable via CLI

### LINKED ISSUE #216: Abnormal CPU usage due to defaultResyncPeriod
## The issue

We were planning to implement the HAProxy Tech Kubernetes Ingress controller for our Kubernetes clusters that are roughly handling the following resources (mileage may vary):

- ~20,000 Ingress
- ~20,000 Service
- ~20,000 Service
- ~20,000 Secret

What we noticed upon installation (although in idle mode and not yet serving traffic), was the really high amount of CPU usage as you can see here from the Grafana snapshot.

Screenshot from 2020-07-09 17-16-33

As it's visible on the graph, the average CPU consumption is more than one vCPU.

Thanks to the `pprof` server offered by the Ingress Controller, I had the chance to figure out the root cause of these spikes.

Screenshot from 2020-07-08 14-26-55

This is the flame graph of a profile session and it's spending ~30% of the CPU time doing garbage collection activities ([profile file is available here for further investigation](https://github.com/haproxytech/kubernetes-ingress/files/4897755/pprof.haproxy-ingress-controller.samples.cpu.001.pb.gz)).

## Root cause

I started digging out the code and noticed this:
https://github.com/haproxytech/kubernetes-ingress/blob/b5889b501b37992a34c7e3b7465280a801ab3c5b/controller/kubernetes.go#L101-L157

The third argument of the `cache.NewInformer` function is the `defaultResyncPeriod`, a trivial parameter that allows setting the time period when the reconciliation between in-memory cache and the controller must be executed: when this happens, all the attached handlers will be called even if an update hasn't been applied, that means all the objects in the in-memory cache will be sent back to the handlers (without retrieving them from the Kubernetes API Server, so no relist action is executed)

Although low values are preferred in order to ensure the desired state is achieved, this is putting so much pressure on the binary since all the objects are going to be processed by the _handler funcs_, escaping from the heap and causing those GC clean-up tasks.

## Proposed solution

I've tested a custom-build using a higher value for the `defaultResyncPeriod` (_30 mins_) and the result is the following:

Screenshot from 2020-07-09 17-32-17

([available profiling output here](https://github.com/haproxytech/kubernetes-ingress/files/4897835/pprof.haproxy-ingress-controller.samples.cpu.009.pb.gz))

Essentially, we're just executing the first reconciliation loop (the cause of the first spikes), and then it calms down except some spikes due to the changes in the production cluster.

Using a value could be tough and not suitable for all the Ingress Controller deployments since there is no one-size-fits-all, so making it dynamic through a CLI option could do the trick.

prometherion:
Closes #216.

Providing the new flag/annotation `--shared-informer-resync-period` with the default value set to 1 minute to make it backward compatible.

Documentation updated, tested locally on my _Kubernetes in Docker_ instance: please, review.

### DIFF
--- a/controller/kubernetes.go
+++ b/controller/kubernetes.go
@@ -101,7 +101,7 @@ func (k *K8s) EventsNamespaces(channel chan *Namespace, stop chan struct{}) {
 	_, controller := cache.NewInformer( // also take a look at NewSharedIndexInformer
 		watchlist,
 		&corev1.Namespace{},
-		1*time.Second, //Duration is int64
+		10*time.Minute, //Duration is int64
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
 				data := obj.(*corev1.Namespace)
@@ -168,7 +168,7 @@ func (k *K8s) EventsEndpoints(channel chan *Endpoints, stop chan struct{}) {
 	_, controller := cache.NewInformer( // also take a look at NewSharedIndexInformer
 		watchlist,
 		&corev1.Endpoints{},
-		1*time.Second, //Duration is int64
+		10*time.Minute, //Duration is int64
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
 				item, err := k.convertToEndpoints(obj, ADDED)
@@ -264,7 +264,7 @@ func (k *K8s) EventsIngresses(channel chan *Ingress, stop chan struct{}) {
 	_, controller := cache.NewInformer( // also take a look at NewSharedIndexInformer
 		watchlist,
 		&extensions.Ingress{},
-		1*time.Second, //Duration is int64
+		10*time.Minute, //Duration is int64
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
 				data := obj.(*extensions.Ingress)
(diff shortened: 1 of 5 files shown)


Mo3m3n:
Thanks for this PR.
Using the `SharedIndexInformers` instead of the low level `Informers` is appreciated, this way no need for separate resources for each Informer. 
However since this also fixes the too low resyncPeriod in #216  I think it is better we do it on separate commits
- First one would be to change the default syncPeriod from 1s to 10mns at least. This can be discussed but I think that there is no need for syncPeriod lower than that, the resync ensures that controller logic is accurate/synced with what's in the Informer cache (by re-triggering object handlers). I'd even say that this can be disabled altogether by default (0 value), because in our case re-triggering handlers will only resync internal representation of k8s objects with what's in Informer cache ([events.go](https://github.com/haproxytech/kubernetes-ingress/blob/master/controller/events.go)) and this will likely lead to nothing (no controller actions) if the object did not change. @oktalz may have more information about this specific part.
- Second commit would be the move to SharedIndexInformers
- Third commit would making this configurable via CLI (can be part of the second though): in this regard I think `shared-informer-resync-period` maybe too cryptic for the average k8s user so maybe something like cache-resync-period is more appropriate.
Also I think first and second commits are no less than MEDIUM severity

prometherion:
@Mo3m3n please, review: just split the changes into 3 commits along with the severity you suggested.

PR #218, opened 2020-07-21
Implementing custom errorfile handlers via dedicated ConfigMap

### LINKED ISSUE #82: Support for errorfile
When some error occur, HAProxy returns by default a "simple" error page, that said, it also allows for customization of this page through the errorfile directive:
https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#errorfile

Would be great to support both globally (defaults section) or per service (backend section) error file for people who want to customize such errors

prometherion:
Closes #82 providing a new CLI flag (`--configmap-errorfile`) using the pattern `<NAMESPACE>/<NAME>`.

# Example of the desired ConfigMap

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
    name: errorfile
    namespace: default
data:
    503.http: |-
        HTTP/1.0 503 Service Unavailable
        Cache-Control: no-cache
        Connection: close
        Content-Type: text/html
    
        <html><body><h1>Oops, that's embarassing!</h1>
        There are no servers available to handle your request.
        </body></html>
```

As mentioned into the docs, the pattern `<code>.http` must be followed as suggested by [the official documentation](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4.2-errorfile).

Many kudos to @Mo3m3n that helped me understanding how the Config Parser deals with `errorfile` type :tada:


Mo3m3n:
Thanks @prometherion for implementing this haproxy feature.
The choice of having the user mount stuff in /usr/local/etc/haproxy/errors in addition to preparing the configMap, seems to be quite cumbersome.
I was rather expecting the user to just provide a configmap like this one:
```  apiVersion: v1
    kind: ConfigMap
    metadata:
      name: errorfile
      namespace: default
    data:
      503.http: |-
        HTTP/1.0 503 Service Unavailable
        Cache-Control: no-cache
        Connection: close
        Content-Type: text/html
    
        <html><body><h1>Oops, that's embarassing!</h1>
        There are no servers available to handle your request.
        </body></html>
```
Then the controller will automatically create a 503 file in `/etc/haproxy/errors/503` and put the HTTP content there.
If that 503.http value is updated/deleted the controller will make the necessary changes.
What do you think ?

Also any reason using 503.http instead of just 503 in the data fields.

prometherion:
Using a ConfigMap provides a huge benefit since Kubelet is able to reload the content upon changes: it means the Controller just needs to load the changed file since Kubelet will ensure the content on the FS is already updated (and since it's an R/O filesystem, cannot be changed).

Regarding the pattern `<code>.http`, as I said, I followed the best practice suggested by the [docs](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4.2-errorfile), as the default location.

> It is recommended to follow the common practice of appending ".http" to
> the filename so that people do not confuse the response with HTML
> error pages, and to use absolute paths, since files are read
> before any chroot is performed.

From my perspective as end-user, mounting a ConfigMap is not a problem: if we decide to go towards this direction, we have obviously to update Helm charts too.

However, if you prefer to avoid the ConfigMap mount (although it's quite convenient) we can easily set a _writer_ to the suggested directory.

prometherion:
@Mo3m3n JFI the latest forced commits are writing errorfile custom responses into `/etc/haproxy/errors` without letting _Kubelet_ to perform file update.

Mo3m3n:
Thanks for the PR 
Work merged in https://github.com/haproxytech/kubernetes-ingress/commit/46abc3935e65ecc36000881e1f1d31da82f7289a

PR #239, opened 2020-10-23
increasing e2e coverage

prometherion:


Mo3m3n:
@prometherion 
You can use the TEST, TEST/MINOR ,etc patch type whenever you work on tests.
Available patch types are in [check-commit](https://github.com/haproxytech/github-actions/blob/7ba9fe614d9412a6f6cb7e18062c166cd2d17e98/check-commit/check.go#L13) action

.github/workflows/kind.yaml
```diff
@@ -17,9 +17,8 @@ jobs:
           name: dev
           config: deploy/kind/kind-config.yaml
           image: kindest/node:${{ matrix.k8s-version }}
-          version: v0.8.1          
+          version: v0.8.1
     - name: Setup
       run: deploy/kind/ci.sh
     - name: Test
       run: go test ./... --tags=integration
```
oktalz, line 24: `\n` at end of the file was removed

oktalz (changes requested):

PR #242, opened 2020-10-30
CI: supporting networking.k8s.io Ingress API group

prometherion:
This PR should fix the CI, ignoring the missing `IngressClass` resource on clusters running version < 1.19.

### DIFF
--- a/controller/kubernetes.go
+++ b/controller/kubernetes.go
@@ -695,8 +695,9 @@ func (k *K8s) IsMatchingSelectedIngressClass(name string) error {
 	}
 	if k.IsNetworkingV1Beta1ApiSupported() {
 		ic, err := k.API.NetworkingV1beta1().IngressClasses().Get(context.Background(), name, metav1.GetOptions{})
-		if err != nil {
-			return fmt.Errorf("the requested IngressClass %s doesn't exist", name)
+		if err != nil && k8serror.IsNotFound(err) {
+			k.Logger.Warningf("the requested IngressClass %s doesn't exist, for upcoming releases this will be mandatory", name)
+			return nil
 		}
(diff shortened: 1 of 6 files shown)


controller/kubernetes.go
```diff
@@ -695,8 +695,9 @@ func (k *K8s) IsMatchingSelectedIngressClass(name string) error {
 	}
 	if k.IsNetworkingV1Beta1ApiSupported() {
 		ic, err := k.API.NetworkingV1beta1().IngressClasses().Get(context.Background(), name, metav1.GetOptions{})
-		if err != nil {
-			return fmt.Errorf("the requested IngressClass %s doesn't exist", name)
+		if err != nil && k8serror.IsNotFound(err) {
+			k.Logger.Warningf("the requested IngressClass %s doesn't exist, for upcoming releases this will be mandatory", name)
```
oktalz, line 699: could we do better her, we know exactly what version this is mandatory on,
So ignore the error prior the specific version 
log warning is OK,. It is not needed because it works on previous versions, but its helpful so upgrade is smoother

why not put a warning in current versions and do an error in versions that already requires it ?
can you even put that here or its complicated ?
because if I'm using `extensions/v1beta1` resource in 1.19, its not and error since that is supported by k8s,
but is `networking/v1beta1` acceptable for k8s to be without ingress class ?

prometherion: The reason behind this is that IngressClass is supported on Kubernetes 1.18, so an end-user would be happy to use it and we should implement it, although it's a sort of mixed behavior: people would use the old approach, others the new one.

This is going to make it smooth, but I say we have to adhere to the Kubernetes best practices, although the previous ones were broken (especially regarding multiple Ingress controller, but that's another story).

PR #253, opened 2020-11-26
Remove lowercasing on path

### LINKED ISSUE #251: Lowercasing of path
I was experimenting with this ingress controller and combining this with the [cert manager](https://cert-manager.io/docs/configuration/acme/http01/).

At first the ingress was created, however the challenge was constantly failing. I managed to determine that the challenge path had uppercase characters and the rules in the `haproxy.cfg` don't allow this at all.

I believe the lowercasing from this rule should be removed, per:

```diff
-http-request set-var(txn.path) path,lower
+http-request set-var(txn.path) path
```

I made my own docker image with this change and was successful in having a certificate issued after this.

I can certainly create a PR to this effect, however I wanted to check if my understanding of this solution is a good one and whether the PR would be accepted first.

ocdi:
I thought I would give master a go to see if it was affected by this issue. I confirmed that certs fail to issue without this change.

The path expressions are case sensitive and fail to match where the ingress has uppercase characters in the request paths. This leaves the path unchanged allowing a correctly cased ingress path to match.

Fixes #251 (on master)

### DIFF
--- a/controller/configuration.go
+++ b/controller/configuration.go
@@ -66,7 +66,7 @@ func (c *Configuration) HAProxyRulesInit() error {
 		c.HAProxyRules.AddRule(rules.ReqSetVar{
 			Name:       "path",
 			Scope:      "txn",
-			Expression: "path,lower",
+			Expression: "path",
 		}, FrontendHTTP, FrontendHTTPS),
 	)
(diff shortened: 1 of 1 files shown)


Mo3m3n:
Thanks @ocdi

Issue #261, opened 2020-12-17
Full reload of haproxy when it shouldn't be necessary


ocdi:
**HAProxy Ingress controller version:**
1.5.0-rc1

**Kubernetes version:**
v1.17.12

**Problem Description**

I noticed when I was playing with master that the ingress controller does a full reload when endpoints are edited. Thought I would try the official `1.5.0-rc1` image to confirm if it wasn't just my compiling but that too has the issue. The silly thing is I can edit an endpoint completely unrelated to the ingress and it does the reload. This is not the behaviour of 1.4.10, on that branch it can add/remove nodes and for this circumstance wouldn't do any change.

While the reload is not a big deal it seems to be a regression compared to the older version.

While modifying an endpoint that is related, it does seem to add the IP using one of the predefined MAINT slots, but then does the full reload anyway.

These are the debug logs output

```
2020-12-17T23:25:09.700359022Z [WARNING] 351/232509 (34) : Server default-web-dev-80/SRV_WfOIG is UP/READY (leaving forced maintenance).
2020-12-17T23:25:09.700484824Z Server default-web-dev-80/SRV_WfOIG is UP/READY (leaving forced maintenance).
2020-12-17T23:25:10.466419732Z 10.96.0.37:15591 [17/Dec/2020:23:25:10.464] stats stats/<STATS> 0/0/0/0/0 200 108975 - - LR-- 4/2/0/0/0 0/0 "GET - HTTP/1.1"
2020-12-17T23:25:11.904967709Z 10.96.0.37:15591 [17/Dec/2020:23:25:11.903] stats stats/<STATS> 0/0/0/0/0 200 108977 - - LR-- 4/2/0/0/0 0/0 "GET - HTTP/1.1"
2020-12-17T23:25:12.591570679Z [WARNING] 351/232512 (34) : Server default-web-dev-80/SRV_WfOIG is DOWN, reason: Layer4 timeout, check duration: 2000ms. 1 active and 0 backup servers left. 0 sessions active, 0 requeued, 0 remaining in queue.
2020-12-17T23:25:12.591570779Z Server default-web-dev-80/SRV_WfOIG is DOWN, reason: Layer4 timeout, check duration: 2000ms. 1 active and 0 backup servers left. 0 sessions active, 0 requeued, 0 remaining in queue.
[...]

PR #262, opened 2020-12-18
Prevent aligning of backends for externalName services

ocdi:
`externalName` services are handled separately and has a single server generated. The `v1.4` branch had this check that appears to have been removed during various refactors. I have tested a custom image with this scenerio and the issue in #261 is no longer present.

In addition to the superfluous reloads, the `externalName` service had an ever growing list of servers appended to the configuration on every reload. This is no longer happening.

### DIFF
--- a/controller/ingress/endpoints.go
+++ b/controller/ingress/endpoints.go
@@ -91,7 +91,9 @@ func (route *Route) handleEndpoints() {
 		return
 	}
 	route.endpoints.BackendName = route.BackendName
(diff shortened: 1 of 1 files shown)


Mo3m3n:
Hi,
Thanks for your help.
Actually I forget to add this to master after I have fixed it on v1.4 https://github.com/haproxytech/kubernetes-ingress/commit/651935b2e56ef8f30c7dca5c010863d09f966bd0

PR #266, opened 2021-01-19
DOC/MEDIUM: Documentation formatting

NickMRamirez:
- Replaced `<pre>` tags with Markdown code blocks. Put code keyword backticks around some keywords. 
- Improved grammar in a few descriptions. 
- Changed the `auth-secret` definition so that the description of the values is under the values key instead of the description key. This makes the table where we show the description only more succinct. 
- For `auth-type` and `auth-secret`, the example now shows both of these keys together because they are dependant upon one another.

### DIFF
--- a/documentation/doc.yaml
+++ b/documentation/doc.yaml
@@ -11,22 +11,24 @@ image_arguments:
         - --configmap=default/my-configmap
   - argument: --configmap-tcp-services
     tip:
-      - Ports of TCP services should be exposed on the controller's kubernetes service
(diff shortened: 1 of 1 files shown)


Mo3m3n:
commit was reverted because 
```
REVERT: commit 2ebd7ef4008fb42d251e9fa8627bfe0c17d5079c
    
    Said commit only updated documentation/doc.yaml but
    documentation/README.md was not updated by running "go run ." from
    documentation/gen. Trying to run it is now giving an error.
```

PR #277, opened 2021-02-23
BUG/MEDIUM: stats frontend: listen IPv6 too

Frankkkkk:
Hi,
I'm not really sure about this MR..
 It adds dual-stack to the stats fronted. However, as-is, it won't obbey to --disable-ipv6 args (who would ever use those anyways ? :P)

Without binding on v6, the stats frontend doesn't work on v6-only clusters,

Let me know what you think.

Cheers !

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -413,6 +413,14 @@ func (c *HAProxyController) handleBinds() (err error) {
 	}
 	if !c.osArgs.DisableIPV6 {
 		protos["v6"] = c.osArgs.IPV6BindAddr
+
+		// IPv6 not disabled, so add v6 listening to stats frontend
+		errors.Add(c.Client.FrontendBindCreate("stats",
+			models.Bind{
+				Name:    "v6",
+				Address: ":::1024",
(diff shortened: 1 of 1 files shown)


Mo3m3n:
Hey @Frankkkkk will this work in v4 only systems ?

Frankkkkk:
Hi,
It does work on v4-only systems, but  it won't work on systems with IPv6 _explicitely_ disabled (`ipv6.disable=1`). I don't know why someone would do that, but at least you know.

To recap: on v4 k8s cluster with a standard config this will work, but not if they also disable the module.

Cheers

oktalz:
@Frankkkkk 
> it won't obbey to --disable-ipv6 args (who would ever use those anyways ? :P)

we added that flag specifically because there are use cases for some that have only ipv4 
and they had issues with ingress controller binding to ipv6.
So while this change is welcomed and can be on by default, it must also go inline and follow same principle as other frontends do
and in case of ` --disable-ipv6` remove that bind

PR #288, opened 2021-03-10
TEST/MINOR: github: checking for generated files diff

prometherion:


.github/workflows/diff.yml
```diff
@@ -0,0 +1,25 @@
...
+on: [push, pull_request]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 10
```
prometherion, line 11: Let's ignore `fetch-depth` if we agree: there's no need for this.

prometherion: Fixed with 7c47dd7

PR #297, opened 2021-04-02
BUG/MINOR: fix whitelist duplication

almson:
Fix for whitelist map file being written to even if it exists.
Not fixed: The fact that the hash of the whitelist is being used for checking equality is really not great. Equality should be checked correctly.

### DIFF
--- a/controller/haproxy/rules.go
+++ b/controller/haproxy/rules.go
@@ -66,7 +66,7 @@ var constLookup = map[RuleType]string{
 type RuleStatus int
 
 // RuleID uniquely identify a HAProxy Rule
-type RuleID uint32
+type RuleID string
 
 //nolint: golint,stylecheck
 const (
@@ -184,7 +184,7 @@ func (r Rules) Refresh(client api.HAProxyClient) (reload bool) {
 					continue
 				}
 				if ftRules.status[id]&INGRESS != 0 {
-					ingressACL = fmt.Sprintf("{ var(%s) -m dom %d }", IngressACLVar, id)
+					ingressACL = fmt.Sprintf("{ var(%s) -m dom %s }", IngressACLVar, id)
 				}
 				err := ruleSet[i].Create(client, &fe, ingressACL)
 				logger.Error(err)
--- a/controller/frontend-annotations.go
+++ b/controller/frontend-annotations.go
@@ -75,15 +75,17 @@ func (c *HAProxyController) handleBlacklisting(ingress *store.Ingress) {
 	}
 	// Validate annotation
 	mapName := "blacklist-" + utils.Hash([]byte(annBlacklist.Value))
-	for _, address := range strings.Split(annBlacklist.Value, ",") {
-		address = strings.TrimSpace(address)
-		if ip := net.ParseIP(address); ip == nil {
-			if _, _, err := net.ParseCIDR(address); err != nil {
-				logger.Errorf("incorrect address '%s' in blacklist annotation in ingress '%s'", address, ingress.Name)
-				continue
+	if !c.cfg.MapFiles.Exists(mapName) {
+		for _, address := range strings.Split(annBlacklist.Value, ",") {
+			address = strings.TrimSpace(address)
+			if ip := net.ParseIP(address); ip == nil {
+				if _, _, err := net.ParseCIDR(address); err != nil {
+					logger.Errorf("incorrect address '%s' in blacklist annotation in ingress '%s'", address, ingress.Name)
+					continue
+				}
 			}
+			c.cfg.MapFiles.AppendRow(mapName, address)
 		}
-		c.cfg.MapFiles.AppendRow(mapName, address)
 	}
 	// Configure annotation
 	logger.Tracef("Ingress %s/%s: Configuring blacklist annotation", ingress.Namespace, ingress.Name)
--- a/controller/frontend-annotations.go
+++ b/controller/frontend-annotations.go
@@ -110,15 +110,17 @@ func (c *HAProxyController) handleWhitelisting(ingress *store.Ingress) {
 	}
 	// Validate annotation
 	mapName := "whitelist-" + strconv.Itoa(int(utils.Hash([]byte(annWhitelist.Value))))
-	for _, address := range strings.Split(annWhitelist.Value, ",") {
-		address = strings.TrimSpace(address)
-		if ip := net.ParseIP(address); ip == nil {
-			if _, _, err := net.ParseCIDR(address); err != nil {
-				logger.Errorf("incorrect address '%s' in whitelist annotation in ingress '%s'", address, ingress.Name)
-				continue
+	if !c.cfg.MapFiles.Exists(mapName) {
+		for _, address := range strings.Split(annWhitelist.Value, ",") {
+			address = strings.TrimSpace(address)
+			if ip := net.ParseIP(address); ip == nil {
+				if _, _, err := net.ParseCIDR(address); err != nil {
+					logger.Errorf("incorrect address '%s' in whitelist annotation in ingress '%s'", address, ingress.Name)
+					continue
+				}
 			}
+			c.cfg.MapFiles.AppendRow(mapName, address)
 		}
-		c.cfg.MapFiles.AppendRow(mapName, address)
 	}
 	// Configure annotation
 	logger.Tracef("Ingress %s/%s: Configuring whitelist annotation", ingress.Namespace, ingress.Name)
--- a/controller/frontend-annotations.go
+++ b/controller/frontend-annotations.go
@@ -74,7 +74,7 @@ func (c *HAProxyController) handleBlacklisting(ingress *store.Ingress) {
 		return
 	}
 	// Validate annotation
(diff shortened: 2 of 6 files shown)


Mo3m3n:
@almson Thanks for this PR.
Can you give a more detailed description of the issue ?
Currently what the controller does is at each sync/control loop it will update controller maps.
But then the maps will be written as haproxy maps only if the content is updated (content hash changes).

almson:
If there are several identical whitelists used by different resources, the current method will add the entries to the same map on each invocation resulting in a map with duplicate entries. The fix is to check that the map has already been created and to leave it alone.

The other issue that I mentioned is that the maps are named and keyed using the hash of their contents. The hash is only 32 bits, so ~100k different whitelists are pretty much guaranteed to hit a collision, and of course given enough users there will be collisions among smaller numbers of whitelists. See https://en.wikipedia.org/wiki/Birthday_problem Switching to 128 bit hash will pretty much solve the problem, although it's still a tad iffy.

almson:
I pushed a fix to the 32-bit hashing

controller/haproxy/rules.go
```diff
@@ -66,7 +66,7 @@ var constLookup = map[RuleType]string{
 type RuleStatus int
 
 // RuleID uniquely identify a HAProxy Rule
-type RuleID uint32
+type RuleID string
```
Mo3m3n, line 69: Changing this requires also updating this line https://github.com/haproxytech/kubernetes-ingress/blob/ebcd3ce6c37494076b877fcaca033368443b671c/controller/haproxy/rules.go#L187
Otherwise it will result in incorrect conversion and buggy haproxy rules selection

Mo3m3n:
> If there are several identical whitelists used by different resources, the current method will add the entries to the same map on each invocation resulting in a map with duplicate entries. 

You are right there will be redundant entries and unnecessary processing in such case, your commit should fix this.

Mo3m3n (commented):
> I pushed a fix to the 32-bit hashing

I was trying to keep hashing size small (thus smaller ruleID size) to get a more readable and maybe faster to process haproxy.cfg.
But  reliability is important so it makes sense to increase this.

almson:
Fixed

Mo3m3n:
Thanks @almson, one last thing: can please do the same for blacklisting ?
It is the same implementation as whitelisting so there will be also an issue of data duplication.
Other than that this looks good to me and it can be merged.

almson:
Done. Also did the same for proxy-protocol.

If I may allow myself to rant, I have to say that the code still has tons of problems in terms of stuff not being deleted and accumulating, including backends, rules, etc. This isn't a slant on the developers, because this code is very complicated with concurrency, reconciliation patterns, and all sorts of "this is how you write a serious controller" extravagance which is hard to pull off. But all of it is unnecessary! All the serious stuff--the reconciliation, concurrency, etc--is already happening inside HAProxy, and does not have to be duplicated in the controller. All this code has to do is write a text file.

You really have to grok the significance of the reconciliation loop already happening in HAProxy. This controller does not have to track what has been added or deleted or the previous value of the config. All it has to do is:

  - Maintain in-memory the *current* state of Kubernetes by receiving change notifications (not HAProxy's state or the previous state)
  - Take out a lock
  - Create in-memory a new HAProxy config (anything that's not needed, like stale backends, won't be added)
  - Write it out to a text file, making sure this is an idempotent operation so that HAProxy doesn't see unnecessary changes
  - Release the lock
 
And then there is no complexity or problems with leaked resources.

Mo3m3n:
> I have to say that the code still has tons of problems in terms of stuff not being deleted and accumulating, including backends, rules, etc.

Do you have examples of this? cause the refresh handler should take care of cleaning the config state from unused backends, rules,etc. If you have examples where this is not happening then it is a bug and it should be fixed.
 
> But all of it is unnecessary! All the serious stuff--the reconciliation, concurrency, etc--is already happening inside HAProxy, and does not have to be duplicated in the controller. 

It would be clearer if we can discuss this based on a specific config/resource where haproxy can handle its reconciliation, concurrency and that is duplicated in controller.

> This controller does not have to track what has been added or deleted or the previous value of the config. All it has to do is:
>  - Maintain in-memory the current state of Kubernetes by receiving change notifications (not HAProxy's state or the previous state)
>  - Take out a lock
>  - Create in-memory a new HAProxy config (anything that's not needed, like stale backends, won't be added)
>  - Write it out to a text file, making sure this is an idempotent operation so that HAProxy doesn't see unnecessary changes
>  - Release the lock

I agree that the controller does not have to track the previous values, actually this is never used in the reconciliation loop and that part of the code need to be cleaned up.
However simplifying the controller by removing code related to tracking what was added/modified/deleted for a given resource and simply recreating configuration when a k8s event is received will result in two annoying drawbacks:
- Unnecessary haproxy reloads
- Loosing any traceability 

Unnecessary haproxy reloads will happen if the controller does not track what exactly changed in a resource. Suppose a kubernetes service was updated and that service is used in an ingress rule. In the design you are describing I guess the controller will update config and reload haproxy even if the updated part is irrelevant like a change in a port name that is not part of the config anyway.
Another cause of unnecessary reloads is when dealing with a change that can applied to haproxy via runtime API (no need for a reload) like backend servers scaling.

Loosing traceability because we are going to update configuration without being able to tell what exactly changed beyond what k8s is telling. For example, currently if an ingress is updated we can precisely log what is being changed in haproxy (which directive in which section of haproxy). If we don't track what changed in a resource, all what we can log is "ingress 'ns/name' updated"

That being said I think that there is room for important improvement, not by completely removing the code tracking what has been changed, but by only  checking the change when receiving k8s events and not dealing with that also in the reconciliation loop via the `store.Status` field which is what we are doing now and bringing additional complexity to the code that we may get rid of. 
I am working on a PR regarding this to see how this would look like

almson:
Avoiding reloads and richer logging are good, but not missing critical. How about this best-effort solution: Make a text diff of the new config and the old. If there's no changes, don't reload. If there's a diff, log it. If the diff shows changes which can be done via API, then do them via API. (Really, HAProxy should be smart enough to do this diff itself and skip reload or make some changes directly without restarting.)

oktalz:
@almson 

> If I may allow myself to rant

this is most definitely welcome, but I see this more as an critique not a rant

 > this code is very complicated with concurrency, reconciliation patterns, and all sorts of "this is how you write a serious controller" extravagance which is hard to pull off.  But all of it is unnecessary! All the serious stuff--the reconciliation, concurrency, etc--is already happening inside HAProxy, and does not have to be duplicated in the controller. All this code has to do is write a text file.
 
 actually, all the changes are done in single loop, gathering data is concurrent.
 Also since changes in k8s are happening constantly, we keep track internally whether or not we acknowledged it or not
 (since we do not reload for every change, but for batch of it). This status can help to decide whether we process resource or not,
 but there are exceptions like http-rules wher ewe trach only if group has changed not a part of it.
 
 and yes, this all can fail (if annotation is configured, of settings have changes on system ... ), so we also take care of this
 
> You really have to grok the significance of the reconciliation loop already happening in HAProxy. This controller does not have to track what has been added or deleted or the previous value of the config. All it has to do is: ...

but it has to, it can't just fire and forget, i HAProxy fails to reload, you end up in state where nothing works anymore.
is up to debate should we just crash also the ingress controller or not, but we currently are more in favor of not crashing controller in that case

Also, it needs to track everything and it needs to only change items that needed to be changed. When having 30k+ (it can really grow) ingress objects, each with few rules, and having the same amount of services (that each might have 1-100+ pods) and writing everything on every k8s change, to me, it does not seem like an efficient way of doing things.

> Maintain in-memory the current state of Kubernetes 

we already do that

> Take out a lock 

its already transactional

>Create in-memory a new HAProxy config 

there is an internal interpretation of config - until saving it you never touch storage

> Write it out to a text file, making sure this is an idempotent operation  

it already works like that

complexity is here because of so many different options and you can switch between them during runtime (like ssl pass-though, certificates, and many other options), no need to kill a controller just to change setup

- can the code be simpler - maybe
- do new options appear that can simplify things - yes (like introducing of maps)
- are we open to discussion about it, yes

almson:
> it needs to track everything and it needs to only change items that needed to be changed... [because] writing everything on every k8s change, to me, it it does not seem like an efficient way of doing things.

Premature optimization is like premature ejaculation. It may be pleasurable, but it doesn't look great.

Have you actually benchmarked how quickly "30k+ (it can really grow) ingress objects, each with few rules, and having the same amount of services (that each might have 1-100+ pods)" can be written out to a file? Because I'll bet it can be done in <100ms, and certainly in <1s.

> but it has to [track changed state], it can't just fire and forget, i HAProxy fails to reload, you end up in state where nothing works anymore

It can just checkpoint the last working state and keep using it until a new config succeeds.

oktalz:
> Premature optimization is like premature ejaculation. It may be pleasurable, but it doesn't look great.

in this case it does look great

> Have you actually benchmarked ...

yes we did

> It can just checkpoint the last working state and keep using it until a new config succeeds.

that is what we do now, if its not changed do not touch it.

Issue #295, opened 2021-03-31
Question: Support of CRD for tcp services


alexandrst88:
Hi! Is there is any plan to manage TCP services via CRDS object or Service annotations instead of putting into the ConfigMap?

I think it would very manageable if for example, each application could define it's own object for tcp load-balancing instead of sharing one Configmap, which could break other services due to human mistake.

oktalz:
Hi, yes there is a plan to add CRDs (includinf tcp services), 
next release 1.6.0 will not use them (most likely), but version after it will have it if everything goes to plan

almson:
Using CRDs to replace ConfigMap for exposing TCP ports would be nice, but perhaps a bit overkill. I hacked together a simpler solution which exposes any annotated NodePorts.

https://github.com/almson/kubernetes-ingress/tree/feature/expose-tcp-nodeports

One of the big issues (upstream as well) is that deleted services don't delete their backends.

PKizzle:
Is this also planned for UDP services? It would be nice if a custom CRD would directly take into account TCP and UDP services.

PR #298, opened 2021-04-02
Feature expose annotated NodePorts

almson:
Replaces the ConfigMap mechanism with examining NodePort annotations directly for deciding how to expose TCP ports.

This isn't a ready, complete pull request. Namely, it's not backwards compatible with ConfigMaps.

See #295

Sample Service:
```
  apiVersion: v1
  kind: Service
  metadata:
    name: myapp
    annotations:
      haproxy.org/expose: 'true'
      haproxy.org/ssl-offloading: 'true'
      haproxy.org/check: 'false'
      haproxy.org/send-proxy-protocol: 'proxy-v2'
      haproxy.org/whitelist: '1.2.3.4/32, 5.6.7.8/28'
  spec:
    selector:
      app: myapp
    type: NodePort
    ports:
    - name: port1
      protocol: TCP
      port: 8023
      nodePort: 30000
    - name: health
      protocol: TCP
      port: 8000
      nodePort: 30001
```

Some issues:
   - Backends don't get deleted
   - Not possible to detect when exposed port changes
   - Not possible to configure SSL individually per port

The reason I'm using NodePorts instead of LoadBalancers is that the latter are hardcoded to create ELBs on AWS (sigh). Also I'm not sure how the exposed port would be specified. (Use nodePort attribute?)

### DIFF
--- a/controller/handler-tcp.go
+++ b/controller/handler-tcp.go
@@ -2,10 +2,12 @@ package controller
 
 import (
 	"fmt"
+	"net"
 	"strconv"
 	"strings"
 
 	"github.com/haproxytech/kubernetes-ingress/controller/haproxy/api"
+	"github.com/haproxytech/kubernetes-ingress/controller/haproxy/rules"
 	"github.com/haproxytech/kubernetes-ingress/controller/store"
 	"github.com/haproxytech/kubernetes-ingress/controller/utils"
 	"github.com/haproxytech/models/v2"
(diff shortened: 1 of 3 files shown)


Mo3m3n:
Hi can you please rebase this PR, I will be able to review it this week

almson:
@Mo3m3n I've rebased, but this isn't a real PR, just wanted to show you how I expose TCP ports based on annotations in NodePort services. I should delete the PR.

Mo3m3n:
> Some issues:
>    Backends don't get deleted
>    Not possible to detect when exposed port changes
>    Not possible to configure SSL individually per port

Thanks for reporting the backends deletion issue, this is actually happening only on dev branch (in stable, backends get deleted correctly) since we have worked lately on some custom routes.
Which means this needs to be fixed globally and not only for tcp servies, which I have pushed in this patch https://github.com/haproxytech/kubernetes-ingress/commit/446846e6c2367589b6ae46d3c6183588fbe014df

> Not possible to configure SSL individually per port

This should be possible with current tcp services [format](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/controller.md#--configmap-tcp-services) where you can enable ssl-offloading per tcp service line.

 >    Not possible to detect when exposed port changes

I agree that moving the tcp service publication from a dedicated configmap to an annotation makes it simpler. However there is no more a dedicated resource/location where all tcp services can be listed. So one has to list all services and check which ones have "expose" annotation. I am note sure how convenient this will be.
Regarding exposed port changes, I think relying on NodePort is a good option to automatically detect changes however this should not be the only way to expose tcp service because:
1. Users will ask how to expose the other type of services (clusterIP, LB, etc)
2. They may also just require that the NodePort and the port exposed via Ingress controller are different which makes sense if you want to publish standard ports but use random ones with NodePorts.

A good tradeoff would be allowing a user to omit the published port in the tcp-services configmap  which will make controller use the NodePort of said service.
What do you think ?

almson:
> A good tradeoff would be allowing a user to omit the published port in the tcp-services configmap which will make controller use the NodePort of said service.
> What do you think ?

I have not thought about and am not proposing a general solution for the controller.

However for my use-case, the problem is having a centralized location of all open ports. This is a problem because I need to launch services and expose ports dynamically. This process should be atomic, while updating a centralized location requires some kind of lock, transaction, etc. So port-per-object (whether that object is a Service or a CRD) works best.

> They may also just require that the NodePort and the port exposed via Ingress controller are different which makes sense if you want to publish standard ports but use random ones with NodePorts.

It looks like v1.21 has a [new feature `spec.loadBalancerClass`](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class) which would support watching LoadBalancer resources instead of NodePorts.

PR #310, opened 2021-05-10
Update config on ConfigMap added event

toshokan:
Currently, if the controller is started without the main ConfigMap existing, but it's added while the controller is running, the running HAProxy config isn't updated, because it's only rebuilt as a result of 'modified' or 'deleted' events, not 'added' events.


Mo3m3n:
Thanks @toshokan 
Can you prepend `BUG/MINOR:` to your commit subject so it follows our patch classification rules

toshokan:
@Mo3m3n done :)

PR #317, opened 2021-05-17
DOC: Generalizes the 'backend-config-snippet' example

NickMRamirez:
Makes the example for `backend-config-snippet` apply to ConfigMap, Ingress and Service. Currently, it applies only to Service.

Note: I had to add some extra whitespace in front of the lines or else it does not render with the correct spacing on haproxy.com/documentation.


Mo3m3n:
Thanks Nick

PR #323, opened 2021-05-26
BUILD: docker: Remove GOOS and GOARCH and use dumb-init from distro

dkorunic:
Stop hardcoding GOOS and GOARCH while building target IC binary, as we need Go
compiler to detect target arch properly.  Also revert to dumb-init from distro
for simpler multiarch images.


build/Dockerfile
```diff
@@ -44,11 +44,9 @@ COPY --from=builder /src/fs/haproxy-ingress-controller .
 
 ARG DUMB_INIT_SHA256=cd7ab5513d20f4b985012d264fbdee60ccd2ea528b94b4b9308c6c6d26f8ba90
```
oktalz, line 45: this is also extra info we do not need anymore

dkorunic: @oktalz Good catch, done in https://github.com/haproxytech/kubernetes-ingress/commit/b05c4c477bf34cda3bffca6752696a21bd12c731.

PR #334, opened 2021-06-03
BUG/MINOR: synchronize servers when Endpoints ports change

fabianonunes:
If a Service (or Endpoint) has its port changed, its servers will be considered unmodified by the controller and their state will not be updated in haproxy.cfg and in `show servers state`. 

Even with the Endpoint port change, the HAProxy will continue to send the requests to the old port.

<details>
<summary>Example</summary>

This Service is used as the default backend and is passed to `--default-backend-service` argument. This bug affects others Services as well.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: default-backend
spec:
  ports:
    - protocol: TCP
      name: http
      port: 1024  # after first deploy, change this port
---
apiVersion: v1
kind: Endpoints
metadata:
  name: default-backend
subsets:
- addresses:
  - ip: 192.168.64.10
  ports:
  - port: 1024 # after first deploy, change this port
```
</details>


ivanmatmati:
Hi @fabianonunes , Thanks for your time and effort really appreciated. I'll give it a review and will keep you inform. Have a nice day.

PR #337, opened 2021-06-07
TCP services listen on IPv6 even when passing --disable-ipv6

fabianonunes:
TCP services always bind on IPv6 and IPv4 despite of `--disable-ipv6` and `--disable-ipv4` flags.

```text
frontend tcp-14031 
  mode tcp
  bind 172.31.10.8:14031 name v4
  bind :::14031 name v6 v4v6  # !!!! this line shouldn't be here,
                              # because `--disable-ipv6` was passed to controller
  option tcplog
  default_backend com-exemplo-echo-1025
```

This PR applies to `TCPServices` the same logic used on `HTTPBind` when creating frontends.

### DIFF
--- a/controller/handler.go
+++ b/controller/handler.go
@@ -45,7 +45,9 @@ func (c *HAProxyController) initHandlers() {
 		handler.TCPServices{
 			SetDefaultService: c.setDefaultService,
 			CertDir:           c.Cfg.Env.FrontendCertDir,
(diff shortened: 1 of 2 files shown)


ivanmatmati:
Hi @fabianonunes, Thanks for this contribution. I'm  going to review it and get back to you if necessary.

oktalz:
Hi @fabianonunes 

thx for the PR, merged

PR #344, opened 2021-06-25
mTLS for tcp frontend (ca-file & verify)

troyanov:
This PR adds support for `client-ca` annotation for tcp frontends
https://www.haproxy.com/documentation/kubernetes/latest/configuration/configmap/#client-ca

Now you can get your frontend with ca-file option and verify required

```
bind *:8883 ssl crt /etc/ssl/crt.pem ca-file /etc/ssl/cacerts.pem verify required
```

### DIFF
--- a/controller/handler/tcp-services.go
+++ b/controller/handler/tcp-services.go
@@ -1,12 +1,14 @@
 package handler
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/haproxytech/client-native/v2/models"
 	config "github.com/haproxytech/kubernetes-ingress/controller/configuration"
+	"github.com/haproxytech/kubernetes-ingress/controller/haproxy"
 	"github.com/haproxytech/kubernetes-ingress/controller/haproxy/api"
 	"github.com/haproxytech/kubernetes-ingress/controller/store"
 	"github.com/haproxytech/kubernetes-ingress/controller/utils"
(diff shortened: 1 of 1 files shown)


ivanmatmati:
Hi @troyanov , thank you for this PR. I've made a review and I'm trying to see how to improve the solution you provided to incorporate the tcp services configmap instead of main one. The implementation of this point could have impact on your code. I'll keep you updated.

troyanov:
> I've made a review and I'm trying to see how to improve the solution you provided to incorporate the tcp services configmap instead of main one.

You mean reading annotation with
https://github.com/haproxytech/kubernetes-ingress/blob/d10c41fb9e7a95322427208b780633797c2e9d8c/controller/handler/tcp-services.go#L35

So for this ConfigMap we can extract `client-ca`:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: tcp
  namespace: default
data:
  "8883": default/vernemq:10883:ssl:client-ca
```

https://github.com/haproxytech/kubernetes-ingress/blob/d10c41fb9e7a95322427208b780633797c2e9d8c/controller/handler/tcp-services.go#L71-L74

I was thinking about it, but got confused with this `:` pattern and how to pass secret name that contains CA file so it can be parsed later on.. 

I can try update my PR to use tcpservice-configmap

ivanmatmati:
This is the idea indeed. But with client-ca and client-crt-optional the configuration line for each tcp service reaches the (psychological) size limit. I'm thinking of a different way to handle tcp services and their configuration with the constraint of preserving as much as possible the current configuration format.

ivanmatmati:
Hi, after discussion the TCP services configuration to accomodate all possible annotations would require an entire refondation as we will certainly do with CRD. Meanwhile, you can go on with the latest proposed solution based on added fields in the tcp service configuration line: "8883": default/vernemq:10883:ssl:client-ca

troyanov:
> Meanwhile, you can go on with the latest proposed solution based on added fields in the tcp service configuration line: "8883": default/vernemq:10883:ssl:client-ca

Thanks for the feedback. Should `client-ca` be a path to a secret containing CA file?

ivanmatmati:
> > Meanwhile, you can go on with the latest proposed solution based on added fields in the tcp service configuration line: "8883": default/vernemq:10883:ssl:client-ca
> 
> Thanks for the feedback. Should `client-ca` be a path to a secret containing CA file?

Yes, exactly. The secret should be in namespace/secret format.

troyanov:
> Any news on this?

Not sure I will ever get back to this.

Issue #364, opened 2021-08-22
make namespace and service name and portname available as variable in config and annotations


nbraun-wolf:
Hi, as I am setting up linkerd with the haproxy controller, I thought it would be great if the namespace service name  and portname could be accessed in config snippets and annotations.

AS example from the[ nginx controller integration documentation](https://linkerd.io/2.10/tasks/using-ingress/#nginx):

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-headers
  namespace: ingress-nginx
data:
  proxy_set_header: "l5d-dst-override $service_name.$namespace.svc.cluster.local:$service_port;"
```

After experimenting for some time I managed to get those three things in the header but unfortunately thei are joined with a hpyhen.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: haproxy
  namespace: ingress
data:
 backend-config-snippet: |
    http-request set-header X-Test %b 
```

So the result looks something like this: `X-Test: demo-app-port-1`. This is already pretty close but not good enough. Because its virtually impossible to try to split this on the hyphens unless we would contrain ourself to not use hyphens in any resource names.

After digging for some time in the codebase, I found out that this `%b` (backend) is set via `%[var(txn.path_match),field(1,.)]`. 

I think 2 options could make sense. 

1. Introduce 3 new variables that contain only the names of namespace, servicename, and portname. Something like `%ns`, `%sn` and `%pn`.
2. Allow to set the delimiter ourselves so that we can use `field` to split those names. i.e. using the tilde `demo~app~port-1`. and then do something like.
```
http-request set-header X-Test %[var(txn.path_match),field(2,-)].%[var(txn.path_match),field(1,-)].cluster.local:%[var(txn.path_match),field(3,-)]
```

PR #367, opened 2021-09-01
change backend names to be able to split it and use the port number

bluebrown:
Use underscore instead of hyphen and always include the port number for backend names.

This will be useful to integrate with linkerd but potentially also with other tools. The indended way to make use of this is by setting an annotation in the global config map.

```yaml
backend-config-snippet: |
  http-request set-header l5d-dst-override %[be_name,field(2,_)].%[be_name,field(1,_)].svc.cluster.local:%[be_name,field(3,_)]
```

#364 

Signed-off-by: Nico Braun <rainbowstack@gmail.com>

### DIFF
--- a/controller/service/service.go
+++ b/controller/service/service.go
@@ -92,9 +92,9 @@ func (s *SvcContext) GetBackendName() (string, error) {
 	}
 	s.path.SvcPortResolved = &svcPort
 	if svcPort.Name != "" {
-		s.backendName = fmt.Sprintf("%s-%s-%s", s.service.Namespace, s.service.Name, svcPort.Name)
+		s.backendName = fmt.Sprintf("%s_%s_%s_%s", s.service.Namespace, s.service.Name, strconv.Itoa(int(svcPort.Port)), svcPort.Name)
 	} else {
-		s.backendName = fmt.Sprintf("%s-%s-%s", s.service.Namespace, s.service.Name, strconv.Itoa(int(svcPort.Port)))
+		s.backendName = fmt.Sprintf("%s_%s_%s", s.service.Namespace, s.service.Name, strconv.Itoa(int(svcPort.Port)))
 	}
 	return s.backendName, nil
 }


bluebrown:
While doing this, I was also thinking, could we not maybe use the FQDN directly as a backend name? 

```go
s.backendName = fmt.Sprintf("%s.%s.svc.cluster.local:%s", s.service.Namespace, s.service.Name, strconv.Itoa(int(svcPort.Port)))
```

Mo3m3n:
I am mitigated about using the service Port number all the time in the backend name.
The problem is that it is possible in the service manifest to change the port number and keep the port name.
For example keep port name `http` but change port number from 80 to 8080.
Now if we're in such case, having the port number in the backend name means deleting and recreating the backend and reloading HAProxy.
While if we use only the port name in the backend name, then the port change can be handled dynamically by updating the port of the backend servers. 
Also it is possible in the future we may need to support different ports in the same backend (multiple server lines with different port numbers) for a given protocol, so in this case it would be better to avoid specifying port number in backend names.  

I can think about two alternatives:
- Providing service ports without port name in service manifests, which will force port number in backend name. But this puts the burden on the user to take care of it.
- Or  smth more reliable like having an annotation for example `backend-name-port-number` that will by default be set to false and the user can set it to true.
In such case the code may look like
```golang
s.backendName = fmt.Sprintf("%s_%s_%s", s.service.Namespace, s.service.Name, svcPort.Name)
usePortNbr, _ = utils.GetBoolValue(input, "backend-name-port-number")
if usePortNbr || svcPort.Name == ""{
    s.backendName = fmt.Sprintf("%s_%s_%s", s.service.Namespace, s.service.Name, strconv.Itoa(int(svcPort.Port)))
}
```

Mo3m3n:
> Another option may be to set another flag set-host-to-fqdn

I agree, it seems setting host to FQDN is a common requirement in service mesh so a dedicated annotation would make sense.
Currently we have already the [set-host](https://github.com/haproxytech/kubernetes-ingress/tree/master/documentation#set-host) so it is possible to have `set-host-to-fqdn` one that internally reuses the set-host logic.  
Now the only thing left is `fqdn-suffix`, I will check first if this is smth the Controller can get automatically from k8s

Issue #373, opened 2021-09-16
Templated string in log line missing call to sprintf 


therealkevinard:
Analyzing pod logs, I found a line that's logging unprocessed string template. 

`xxx` is my redaction; `Ingress %s/%s` is actual log line. 

```
2021/09/16 18:25:34 TRACE   status/ingress.go:130 Successful update of LoadBalancer status in ingress xxx/xxx-sub-xxx
2021/09/16 18:25:34 TRACE   status/ingress.go:82 Updating status of Ingress %s/%s                         <------------[ %s/%s ]
2021/09/16 18:25:34 TRACE   status/ingress.go:82 xxx
```

(MR coming)

PR #374, opened 2021-09-16
FIX/SMALL: process template strings in "updating status of ingress..." log 

therealkevinard:
this passes the arguments through fmt.Sprintf to create meaningful log line. 

resolves https://github.com/haproxytech/kubernetes-ingress/issues/373


therealkevinard:
(hold) fixing commit message per contrib notes.

therealkevinard:
> hi @therealkevinard,
> 
> thx for this,
> but I believe there is also `logger.Tracef` that is more appropriate for fixing this error

🤦‍♂️ Yep. I'll push that up tomorrow.

Mo3m3n:
You need to remove the first commit: the one adding sprintf.
Also better keep the `BUG` patch type rather than `REFACTOR`

Issue #375, opened 2021-09-17
Documentation: 'scale-server-slots' requires quotes around value


NickMRamirez:
The documentation should be updated to show that the `scale-server-slots` annotation requires quotes around its value.

PR #376, opened 2021-09-17
DOC: Added quotes to 'scale-server-slots' value.

NickMRamirez:
Fixes https://github.com/haproxytech/kubernetes-ingress/issues/375


NickMRamirez:
Replaced with https://github.com/haproxytech/kubernetes-ingress/pull/378

PR #413, opened 2022-01-14
BUG/MINOR: sync cm name with config argument

dosmanak:
fixes 5eace95 namespace name change

### DIFF
--- a/deploy/haproxy-ingress-daemonset.yaml
+++ b/deploy/haproxy-ingress-daemonset.yaml
@@ -165,7 +165,7 @@ spec:
         image: haproxytech/kubernetes-ingress
         args:
           - --configmap=haproxy-controller/haproxy-kubernetes-ingress
-          - --default-backend-service=default/haproxy-kubernetes-ingress-default-backend
+          - --default-backend-service=haproxy-controller/haproxy-kubernetes-ingress-default-backend
         securityContext:
           runAsUser:  1000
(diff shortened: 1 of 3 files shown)


Mo3m3n:
Thanks @dosmanak 
In the same context,  path of default-backend-service needs to be fixed too:
https://github.com/haproxytech/kubernetes-ingress/blob/1966e37d7315be071fe15c4321fce2f9aaf5578e/deploy/haproxy-ingress-daemonset.yaml#L168
https://github.com/haproxytech/kubernetes-ingress/blob/1966e37d7315be071fe15c4321fce2f9aaf5578e/deploy/haproxy-ingress.yaml#L169

And this one 
https://github.com/haproxytech/kubernetes-ingress/blob/1966e37d7315be071fe15c4321fce2f9aaf5578e/deploy/tests/rebuild.sh#L22

dosmanak:
Yeah, you're right. I have in my kustomize file and forgot to put it here. Thank you for pointing that out. I've added second commit.

Mo3m3n:
Only this one left (replace default -> haproxy-controller) as already mentioned  and this is good to be merged 
https://github.com/haproxytech/kubernetes-ingress/blob/1966e37d7315be071fe15c4321fce2f9aaf5578e/deploy/tests/rebuild.sh#L22

Mo3m3n:
Thanks @dosmanak

PR #417, opened 2022-01-31
MINOR: Use patternfile in whitelist/blacklist

thriqon:
As a special case, if whitelist or blacklist start with `patterns/`, use this map directly. This enables centrally managing IP ranges and using them in many Ingresses. For example:

```yaml
--                     
kind: ConfigMap                  
apiVersion: v1             
metadata:                
  name: patternfiles             
  namespace: haproxy-controller
data:
  ips: |     
    192.168.0.0/24
---      
kind: Ingress    
metadata:
  name: http-restricted-echo
  annotations:     
    ingress.class: haproxy
    haproxy.org/whitelist: patterns/ips
spec:            
  rules:           
    - host: localhost
      http:            
        paths:
          - path: /restricted
            pathType: Prefix
            backend:
              service:          
                name: http-echo
                port:
                  name: http
```

----

**NOTE** I've skipped writing documentation and more refactoring for now. I first want to get feedback if it is worth investing more time.


thriqon:
@Mo3m3n , I've added an E2e test and documentation.

Mo3m3n:
@thriqon  Perfect work (commit, tests, doc, etc) thanks !
My only remark is about having the accesscontrol e2e test as `e2e_sequential`.
Actually when the test execution and validation won't impact other tests we usually put it as `e2e_parallel` this way the test suite may run a little more faster. 
Tests run sequentially are only tests which may impact the whole Ingress Controller like testing a configmap option (tls-auth for example) or tests that require a single change at a time like endpoints tests where we chuck that the IC did not reload.
So I think the accesscontrol test can go under `e2e_parallel`

Mo3m3n:
I see because patternfiles are also used in the `haproxy-files` tests (which are run in parallel mode) so if these tests are run at the same time as `access-control` ones we will have conflicts.
Ok I am merging this now and will handle this properly later, there are two options I can think about:
- Sharing the same patternfile config between these tests so it will include all required data (pattern files).
- Find a way to append data to the existing patternfiles instead of overriding it.

PR #425, opened 2022-03-02
Update controller.md for https bind port

kalyanteja19:


Mo3m3n:
@kalyanteja19 you will need to follow the subject commit [format](https://github.com/haproxy/haproxy/blob/master/CONTRIBUTING#L632) we use in haproxy projects in this case you just need to prepend your commit message with `DOC/MINOR:`

PR #435, opened 2022-04-05
Use slice instead of static sized arrays.

danlenar:
Since whitelisted namespaces were being appended anyways, the first entries in the
namespaces were all empty strings thus causing ingress controller to access
k8s resources on a cluster level unnecessary.


ivanmatmati:
Hi @danlenar , thanks for this PR. Could you add the expected tags in your commit message ? I guess it should be "BUG/MINOR". You can fin example of commit messages [here](https://github.com/haproxy/haproxy/blob/master/CONTRIBUTING#L846).

danlenar:
@ivanmatmati Added the tag BUG/MINOR

PR #446, opened 2022-05-28
MINOR: allow custom ports for local peer, stats and healthz

### LINKED ISSUE #348: (Multiple ingress same node) vs (stats port 1024 and healthz port 1042 )
The network mode of ingress in this scenario is : useHostNetwork: true
The stats port 1024 and healthz port 1042 are binding in ip 0.0.0.0 ...  so if  I have more than one ingress controller in the same node it does not work ...
How can I bind the stats port and healthz port to an specific ip ? (just like the http and https ports)
I realized that there is a third uncontrolled port  : 127.0.0.1: 10000
So this 3 ports (0.0.0.0:1024 0.0.0.0:1042 127.0.0.1:10000) do  not allow multiple ingress in  the same nodes with useHostNetwork: true.
The nginx  ingress controller allow this configuration. I was using it when I decided to migrate to haproxy.
Thanks  a lot in advance.

fabianonunes:
Currently, it is not possible to simultaneously run two **host-mode** ingress controllers on the same node, because the binds `127.0.0.1:10000` ([local peer](https://github.com/haproxytech/kubernetes-ingress/blob/9e4371177f86bc48c63e79c56968a4f459666ae4/fs/usr/local/etc/haproxy/haproxy.cfg#L17)), `*:1024` ([stats](https://github.com/haproxytech/kubernetes-ingress/blob/9e4371177f86bc48c63e79c56968a4f459666ae4/fs/usr/local/etc/haproxy/haproxy.cfg#L39)), and `0.0.0.0:1042` ([health](https://github.com/haproxytech/kubernetes-ingress/blob/9e4371177f86bc48c63e79c56968a4f459666ae4/fs/usr/local/etc/haproxy/haproxy.cfg#L32)) are hardcoded.

This PR allows using custom ports for local peer, stats and healthz.

Fixes #348

---

Considering some minor API changes between 1.7 and master, I backported this change to 1.6 and 1.7 branches as well (#448, #447)

### DIFF
--- a/fs/usr/local/etc/haproxy/haproxy.cfg
+++ b/fs/usr/local/etc/haproxy/haproxy.cfg
@@ -36,7 +36,7 @@ frontend healthz
 
 frontend stats
    mode http
-   bind *:1024
+   bind *:1024 name stats
    http-request set-var(txn.base) base
    http-request use-service prometheus-exporter if { path /metrics }
    stats enable
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -171,13 +171,14 @@ func (c *HAProxyController) updateHAProxy() {
 
 // setToRready exposes readiness endpoint
 func (c *HAProxyController) setToReady() {
+	healthzPort := c.osArgs.HealthzBindPort
 	logger.Panic(c.clientAPIClosure(func() error {
 		return c.haproxy.FrontendBindEdit("healthz",
 			models.Bind{
 				BindParams: models.BindParams{
 					Name: "v4",
 				},
-				Address: "0.0.0.0:1042",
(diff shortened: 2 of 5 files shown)


fabianonunes:
I can't figure out why the last check failed. On my fork, the same commit with the same parent [ran successfully](https://github.com/fabianonunes/kubernetes-ingress/runs/6637746276?check_suite_focus=true).

fabianonunes:
I can't figure out why the last check failed. On my fork, the same commit with the same parent [ran successfully](https://github.com/fabianonunes/kubernetes-ingress/runs/6638165655?check_suite_focus=true).

oktalz:
hi @fabianonunes thx !

PR #448, opened 2022-05-28
Allow custom ports for local peer, stats and healthz (1.7 backport)

### LINKED ISSUE #348: (Multiple ingress same node) vs (stats port 1024 and healthz port 1042 )

fabianonunes:
This PR backports #446 to v1.7 branch.

----

Currently, it is not possible to simultaneously run two host-mode ingress controllers on the same node, because the binds `127.0.0.1:10000` (local peer), `*:1024` (stats), and `0.0.0.0:1042` (healthz) are hardcoded.

This PR allows using custom ports for local peer, stats and healthz.

Fixes #348

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -15,6 +15,7 @@
 package controller
 
 import (
+	"fmt"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


fabianonunes:
I can't figure out why the last check failed. On my fork, the same commit with the same parent [ran successfully](https://github.com/fabianonunes/kubernetes-ingress/runs/6638153545?check_suite_focus=true).

PR #451, opened 2022-06-02
BUG/MINOR: using more than one syslog server always leads to unnecessary restarts

### LINKED ISSUE #420: controller.go:205 HAProxy restarted (1.7.4)
Running 1.6.8 (and 1.6.x series) and things fine
Upgrading to 1.7.4 and see instability and substantial numbers of
`controller.go:205 HAProxy restarted`
Nothing changed in cluster other than upgrade.
Downgrading back to 1.6.8 resolves.

Any suggestions as to what source of instability might be ?
Many Thanks

fabianonunes:
Fixes #420

When using more than one syslog server, the algorithm for comparing log targets never succeeds, because the conversion of each `syslog-server` annotation's line into a `models.LogTarget` will always set the `.Index` field equal to 0:

https://github.com/haproxytech/kubernetes-ingress/blob/88a9cb4a106a084c93549909f9f862afeca4de85/pkg/annotations/global/syslogServer.go#L59

In this scenario, any state change causes a restart because `lg` and `newLg` will never be equal (either because of the slice order, or because of the different `.Index` fields):

https://github.com/haproxytech/kubernetes-ingress/blob/88a9cb4a106a084c93549909f9f862afeca4de85/pkg/controller/global.go#L84-L89

## Example

ConfigMap input:
```yaml
syslog-server: |
  address:stdout, format:raw, facility:daemon, level:notice
  address:/run/shared/log, facility:local0, level:info
```

Controller output after a simple scale-up (formatted, for easier reading):
```
2022/06/02 00:41:59 DEBUG   global.go:82 Global log targets updated:
[
slice[0].Address: stdout != /run/shared/log
slice[0].Facility: daemon != local0
slice[0].Format: raw != 
slice[0].Level: notice != info

slice[1].Address: /run/shared/log != stdout
slice[1].Facility: local0 != daemon
slice[1].Format:  != raw
slice[1].Index: 0 != 1
slice[1].Level: info != notice
]
Restart required
```

### DIFF
--- a/pkg/annotations/global/syslogServer.go
+++ b/pkg/annotations/global/syslogServer.go
@@ -35,7 +35,7 @@ func (a *SyslogServers) GetName() string {
 func (a *SyslogServers) Process(k store.K8s, annotations ...map[string]string) error {
 	input := common.GetValue(a.GetName(), annotations...)
 	a.stdout = false
-	for _, syslogLine := range strings.Split(input, "\n") {
+	for i, syslogLine := range strings.Split(input, "\n") {
 		if syslogLine == "" {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


fabianonunes:
I can't figure out why the last check failed. On my fork, the same commit with the same parent [ran successfully](https://github.com/fabianonunes/kubernetes-ingress/runs/6702031334).

ivanmatmati (approved):

oktalz:
thx!

PR #453, opened 2022-06-04
MINOR: copies binary last, to favor caching and speed up the image build

fabianonunes:
This PR reorders image layers, copying binary last to favor caching and speed up the image build.


oktalz:
thx!

PR #457, opened 2022-06-05
BUG/MEDIUM: Merge HTTPConnectionMode annotations into one.

fabianonunes:
Both the `http-keep-alive` and `http-server-close` annotations change the value of `defaults.HTTPConnectionMode`. But since the `http-keep-alive` annotation is evaluated last, it overwrites whatever value `http-server-close` has previously set, even if `http-keep-alive` is empty, undefined, or `'false'`

Perhaps it might be better to merge these annotations into one since they are mutually exclusive. After all, if you set `http-keep-alive: 'false'`, you won't get the expected result: no options will be set and the default will be used, **which is keep-alive**. Also, a third alternative is not supported via annotations ([`httpclose`](https://cbonte.github.io/haproxy-dconv/2.5/configuration.html#4.2-option%20httpclose)).

So, I think it might be better to have an `http-connection-mode` annotation, whose values could be: `http-keep-alive` (default), `http-server-close` or `httpclose`


oktalz:
hi @fabianonunes thx for this.

it is a good idea, however we can't remove existing annotations since that would break existing setups.
what we can do is to deprecate them (but still support them).

if they are used, we can issue an log warning that this will be removed, but they still must work for a certain time.
(one or two next IC versions, even more)

So, old behavior must still work if user was setting those annotations, 
but if they do not exist, new one is used.

that could be handeled in `pkg/annotations/global/connectionMode.go` that you created

fabianonunes:
Hi, @oktalz!

This is a much better idea!!! I will adjust this PR.

fabianonunes:
Hi, @oktalz.

I followed your advice and dealt with the old annotations in `pkg/annotations/global/connectionMode.go`. To keep the changes in just one commit, I force-pushed on the same branch (I don't know if this was a good idea).

Rather than putting the logic inside the `HTTPConnectionMode.Process` function, I preferred extracting it into another method, making it easier to remove it when the old annotations become obsolete.

After these changes, this PR fixes the bug reported on #450, introduces a new annotation, and keeps the old ones so the old behavior still works. I've adjusted the docs too.

With the new handling of these annotations, we have the following scenarios:

```text
o http-server-close:    "true"
o http-keep-alive:      "false" or empty or unset
o http-connection-mode: empty or unset

> HTTPConnectionMode = http-server-close
```
```text
o http-server-close:    "true"
o http-keep-alive:      "true"
o http-connection-mode: empty or unset


> HTTPConnectionMode = http-keep-alive
- keep-alive takes precedence over server-close
```

```text
o http-server-close:    "false" or empty or unset
o http-keep-alive:      "false" or empty or unset
o http-connection-mode: empty or unset

> HTTPConnectionMode = http-keep-alive
- when all are falsy, haproxy default value is used
```

```text
o http-server-close:    "true"
o http-keep-alive:      "sí" # invalid value
o http-connection-mode: empty or unset

> HTTPConnectionMode = http-server-close
- old annotations with invalid values are ignored but the error is reported
```

```text
o http-server-close:    "true" or "false" or empty or unset
o http-keep-alive:      "true" or "false" or empty or unset
o http-connection-mode: "http-keep-alive" or "http-server-close" or "httpclose"

HTTPConnectionMode = the same as `http-connection-mode`
- when set, http-connection-mode takes precedence over the old annotations
```

fabianonunes:

ivanmatmati (approved):

PR #462, opened 2022-06-15
BUG/MINOR: Prevents unnecessary reloads when `cookie-persistence` is enabled.

### LINKED ISSUE #459: Frequent reloads with cookie-persistence enabled
The is with version 1.7.9. We noticed frequent reloads happening multiple times a minute, which we couldn't explain.
After turning on debug logging, we see messages looking like this, that seem to show the cause

```
DEBUG   service/service.go:122 Service 'redacted/redacted': backend 'redacted' updated: [Cookie.Domains: <nil slice> != []]
Reload required
```
We have cookie persistence enabled on the ingress with the annotation

```
haproxy.org/cookie-persistence: COOKIE
```

this results in a generated configuration with a cookie line this

```
 cookie COOKIE dynamic indirect nocache insert 
```

It seems like maybe a type bug when comparing the expected and running configuration for the cookie settings?

fabianonunes:
Fixes #459 

When a Service/Ingress/ConfigMap is configured with the annotation `cookie-persistence`, the comparison with the corresponding backend in the HAProxy settings will always fail, because the `backend.Cookie.Domain` attribute of an initialized model is a `<nil slice>` and the corresponding value returned by HAProxy is an empty array.

```text
service/service.go:134 Service 'exemplo/bazaar': backend 'exemplo_bazaar_http' updated: [Cookie.Domains: <nil slice> != []]
Reload required
```

### DIFF
--- a/pkg/annotations/service/cookie.go
+++ b/pkg/annotations/service/cookie.go
@@ -36,6 +36,7 @@ func (a *Cookie) Process(k store.K8s, annotations ...map[string]string) error {
 		Nocache:  true,
 		Indirect: true,
 		Dynamic:  true,
+		Domains:  []*models.Domain{},
 	}
 	return nil
 }


fabianonunes:

ivanmatmati (approved):

PR #500, opened 2022-11-10
MINOR: Update dep with vulnerabilities

choffmeister:
Updated golang.org/x/text dependency to fix CVE.

CVE-2022-32149
GHSA-69ch-w2m2-3vjp


oktalz:
@choffmeister thx for this, however there are some issues with this.
- commit message indicates that you are removing and updating deps, but you are adding `ca-certificates`
- its not necessary to add `ca-certificates` at all (its already there)
- removing curl is not desired, it is a useful tool that we plan to keep.
  - however we might think about different solutions, ideally alpine updates curl version and that is it
- updating `golang.org/x/text` is desired and for that I can accept PR

PR #515, opened 2023-01-12
BUG/MINOR: log error when multiple default IngressClasses found

yasinlachiny:
Ensuring that at most 1 IngressClass is marked as default in your cluster otherwise show a log error based on it.

### DIFF
--- a/pkg/ingress/ingress.go
+++ b/pkg/ingress/ingress.go
@@ -59,14 +59,17 @@ func (i Ingress) supported(k8s store.K8s, a annotations.Annotations) (supported 
 		igClassSpec = igClassResource.Controller
 	}
 	if igClassAnn == "" && i.resource.Class == "" {
+		var defaultClassFound bool
 		for _, ingressClass := range k8s.IngressClasses {
 			if ingressClass.Annotations["ingressclass.kubernetes.io/is-default-class"] == "true" {
+				if defaultClassFound {
+					logger.Errorf("multiple default IngressClasses found")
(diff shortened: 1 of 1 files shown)


oktalz:
@yasinlachiny thanks for the PR.

I have few observations:
- its being run for every ingress object
  - that would add significant slowdown in some cases. Imagine having 20k+ ingresses (some users have that)
  - in case of many ingresses we would spam logs with same message
- from log message its not clear what is in conflict
  - it would be better if instead of `multiple default IngressClasses found` we have `multiple default IngressClasses found: [name1, name2]`
  - log message should be warning, not error (controller still works)
  
  the best place for that check is in [events](https://github.com/haproxytech/kubernetes-ingress/blob/master/pkg/store/events.go#L41) where this can be done once per ingress class event (we can even ignore check on delete)

yasinlachiny:
Hi @oktalz 
Thank you for reviewing my code.

- It loops over ingress classes not ingress and we have a limited number of ingress classes. A company can surely have thousands of ingresses but they have a few ingress classes.

- Yeah. A warning is better, I'll change it.

oktalz:
> It loops over ingress classes not ingress and we have a limited number of ingress classes

yes, it is definitely expected to have few ingress classes.
But if its even one class, it its being done over **every** ingress, `function (i Ingress) supported(` is called for every ingress.
and if we say N is number of ingresses, that is being done N times, instead of once in `EventIngressClass` function

Issue #318, opened 2021-05-20
Used proxy-protocol for TCP Service


tili41:
Hi, we have an external HAProxy load-balancer upstream of the K8s cluster.
We have enabled the proxy-protocol to be able to whitelist the ip allowed to connect to the different services managed by HAProxy Controller. It works well for HTTPS services. 
However, it doesn't seem possible to enable the proxy-protocol for TCP services.

Without the proxy-protocol, I do not really have the interest to have whitelists for TCP services (we manage this via the Service backend-config-snippet annotation). 
To be clear, we don't want to enable proxy-protocol for the backend servers but only for the frontend (Ingress Controller).
Any idea? Thank you for your help.

tili41:
Hi. Do you have any news about this? Possible or not? If not, in a later release? Thanks.

ivanmatmati:
Hi @tili41, Currently this is not available on TCP services. We could plan it for a future release.

Mo3m3n:
I think we better enable [config-snippets](https://github.com/haproxytech/kubernetes-ingress/tree/master/documentation#config-snippet) for TCP frontends, this way you can add whatever configuration you want.

apanagiotou:
Any news on making proxy-protocol available for TCP services?

strowi:
Any update on this? Or an example on how to enable accept-proxy on the frontend?

vholer:
Also interested in an easy way how to enable proxy for other frontends (e.g., stats).

PR #525, opened 2023-03-01
Increase options for tcp services. Add accept-proxy to the list of available parameters with backward compat.

ctrix:
At $dayjob we use haproxy on TCP services and we rely on the proxy protocol to transport the remote endpoint ip address to the backends.

Very often we chain several proxies one in front of the other. We noticed that the ingress controller doesn't handle the accept-proxy option, while it is supported for HTTP(s) connections through other settings.

This patch reuses the third parameter of the tcp services list, previously optional and supporting only `ssl` as value and transforms it into a comma separated list of options; It also adds the `accept-proxy` option support.

When the tcp service will contain `accept-proxy` in the options, it will be added to the bind line in the configuration file.

This should also address several requests and issues, like #318 and should enable more features to be added to the tcp services keeping the backward compatibility.


ctrix:
Done.

oktalz:
hi @ctrix 

currently we plan to expand capabilities for TCP services with gateway API.
because of that we need to put this on hold for a while to determine what is missing there.
in any case expanding options here would bring a bit of confusion

since gateway api is still in alpha/beta and it is uncertain when it will be finalized form k8s team, 
a better option would be to create a CRD for TCP services that would replace a bit awkward structure that we can have with configmaps.

Because of that I will put this on hold for now until situation is a bit more clear

ShadowJonathan:
@oktalz has there been any movement in this?

oktalz:
@ShadowJonathan yes, tomorrow we will release v1.11 and next release after that one contains TCP CRD in roadmap.

structure and options in it will probably be added gradually (over development time) and I'll notify here when changes are done so we can align with expected options

Raj-prateek:
Hi @oktalz , Do you have any updates on TCP CRD release?

oktalz:
@ctrix, @Raj-prateek, @ShadowJonathan 
definitions are here, and usage is explained here [documentation/custom-resource-tcp.md](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/custom-resource-tcp.md)

please take a look and see if this approach will be something you can use, please note that options for tcp CRD will expand in future

example in doc shows how you can use this for `accept-proxy` option

in general we opted for full option set of configuration options, its very similar to what you have in dataplaneapi for example, in fact we use same scheme (here its just merged in k8s way)

oktalz:
due to some changes in behavior of IC (accepting more options), tcp crd will be available only in new version 3.0 (to align version with underlying HAProxy version) and new version is planned withing next two weeks (possibly even this week)

clement0210:
Hello @oktalz,
Do you have an updated ETA for this release please?

hdurand0710:
Hello @clement0210 ,
It should be released this week.

oktalz:
I'm going to close this PR since we have [TCP CRD](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/custom-resource-tcp.md) now.
In any case I really appreciate this and a push for them.

If anything is missing, we will add it to TCP CRD since it is a better solution than using TCP configmap for this.

Issue #519, opened 2023-01-20
The error message shows 'blacklist' for whitelist annotation


dosmanak:
https://github.com/haproxytech/kubernetes-ingress/blob/d772b9db117e086d1affe7cbb3111ebb5cd0618e/pkg/annotations/ingress/accessControl.go#L61

ivanmatmati:
A fix has been merged.

PR #529, opened 2023-03-09
Remove unecessary information from whitelist/blacklist address format…

benm-stm:
Corrects issue #519 
here is the output of the error:
annotation blacklist: incorrect address '192.168.1.0-24' in blacklist annotation'
annotation whitelist: incorrect address '192.168.1.0-24' in blacklist annotation'

We already have the information in the beginning of the error

### DIFF
--- a/pkg/annotations/ingress/accessControl.go
+++ b/pkg/annotations/ingress/accessControl.go
@@ -58,7 +58,7 @@ func (a *AccessControl) Process(k store.K8s, annotations ...map[string]string) (
 			address = strings.TrimSpace(address)
 			if ip := net.ParseIP(address); ip == nil {
 				if _, _, err := net.ParseCIDR(address); err != nil {
-					return fmt.Errorf("incorrect address '%s' in blacklist annotation'", address)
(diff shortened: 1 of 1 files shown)


pkg/annotations/ingress/accessControl.go
```diff
@@ -58,7 +58,7 @@ func (a *AccessControl) Process(k store.K8s, annotations ...map[string]string) (
 			address = strings.TrimSpace(address)
 			if ip := net.ParseIP(address); ip == nil {
 				if _, _, err := net.ParseCIDR(address); err != nil {
-					return fmt.Errorf("incorrect address '%s' in blacklist annotation'", address)
+					return fmt.Errorf("incorrect address '%s'", address)
```
ivanmatmati, line 61: Maybe you can keep the origin of the issue with
```suggestion
					return fmt.Errorf("incorrect address '%s' in %s annotation", address, a.name)
```

ivanmatmati:
Hi, a last thing would be to squash your two commits in a single one. Thanks.

oktalz:
hi @benm-stm,

thx for the PR!

Issue #530, opened 2023-03-09
FEAT/ Apply TCP mode loadbalance algorithm  to services


benm-stm:
Add a loadbalancing algorithm in the tcp-configmap this way
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubernetes-ingress
  namespace: default
  annotations:
    haproxytech/load-balance: leastconn
data:
  "1025": namespace/svc:8080
```
This latter one will be applied to all tcp services in the tcp-configmap

benm-stm:
Or better, imbed load-balance algorithm in the data section per service
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubernetes-ingress
  namespace: default
data:
  "1025": namespace/svc:8080:load-balance=leastconn
  "1026": namespace/svc:8081:load-balance=roundrobin
```

ivanmatmati:
Response sent directly in an other canal.

PR #533, opened 2023-03-21
feat:#530 add tcp services annotation capability

benm-stm:
I worked with tcp services defined in configmaps, one blocking point was the lack of annotations
I tried to change load-balance algorithms for every service
so i created this contribution
here is what we can do now
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kubernetes-ingress
  namespace: default
data:
  "1025": namespace/svc:8080:ssl:load-balance=leastconn,check=true
  "1026": namespace/svc:8081::load-balance=roundrobin
```
the last field can be any of the valid annotations
I created a unit test too to validate the usecase

### DIFF
--- a/pkg/handler/tcp-services.go
+++ b/pkg/handler/tcp-services.go
@@ -25,6 +25,7 @@ type tcpSvcParser struct {
 	service    *store.Service
 	port       int64
 	sslOffload bool
+	annList    map[string]string
 }
 
 func (handler TCPServices) Update(k store.K8s, h haproxy.HAProxy, a annotations.Annotations) (reload bool, err error) {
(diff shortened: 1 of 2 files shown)


oktalz:
hi @benm-stm,

currently we plan to expand capabilities for TCP services with gateway API.
because of that we need to put this on hold for a while to determine what is missing there.
in any case expanding options here would bring a bit of confusion, now we have `::` and `:` and `,` to distinct all options.

since gateway api is still in alpha/beta and it is uncertain when it will be finalized form k8s team, 
a better option would be to create a CRD for TCP services that would replace a bit awkward structure that we can have with configmaps.

Because of that I will put thig on hold for now until situation is a bit more clear

oktalz:
hi @benm-stm 

TCP CRD definitions are here, and usage is explained here [documentation/custom-resource-tcp.md](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/custom-resource-tcp.md)

please take a look and see if this approach will be something you can use, please note that options for tcp CRD will expand in future.

in general we opted for full option set of configuration options, its very similar to what you have in dataplaneapi for example, in fact we use same scheme (here its just merged in k8s way),
so there won't be any need to add additional annotations, all options should be available in CRD

oktalz:
I'm going to close this PR since we have TCP.
In any case I really appreciate this and a push for them.

If anything is missing, we will add it to TCP CRD since it is a better solution than using TCP configmap for this.

PR #536, opened 2023-04-19
BUG/MEDIUM: fix typo in healthcheck endpoint

thecodingrobot:
By default the aws load balancer controller expects the path to be /healthz Ref.: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/service/annotations/#health-check

### DIFF
--- a/pkg/controller/builder.go
+++ b/pkg/controller/builder.go
@@ -202,7 +202,7 @@ func addControllerMetricData(builder *Builder, chShutdown chan struct{}) {
 		rtr.GET("/metrics", fasthttpadaptor.NewFastHTTPHandler(promhttp.Handler()))
 		runningServices += ", prometheus"
 	}
-	rtr.GET("/healtz", requestHandler)
+	rtr.GET("/healthz", requestHandler)
 	// all others will be 404
 	go func() {
 		server := fasthttp.Server{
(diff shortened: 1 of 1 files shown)


ivanmatmati:
Hi @thecodingrobot , Thanks for the PR. We'll process it soon.

oktalz:
@thecodingrobot thx for this

could you leave existing endpoint (just add a correct one before it)

we will deprecate faulty one, 
but since I can assume that this is already used somewhere, its better to keep it there for version or two and later remove it, to allow people to switch without immediate breakage of existing deployments

eifelmicha:
Is there any plan to go forward on this?

oktalz:
@eifelmicha I though my comment was clear, we will not introduce breaking change in proposed manner.

instead of 
```
- rtr.GET("/healtz", requestHandler)
+ rtr.GET("/healthz", requestHandler)
```

we need just
```
+ rtr.GET("/healthz", requestHandler)
```

if the original author does not change it, we will add in next few weeks

eifelmicha:
@oktalz 
Your comment is quite clear. I tested HAProy coming from https://github.com/jcmoraisjr/haproxy-ingress i noticed difficulties like this and wondered if they will be fixed at any time. Would like to use this upstream haproxy.

oktalz:
@eifelmicha if @thecodingrobot does not respond in next few days, we will add different commit that does not break current behavior, but still fixes mentioned issue

ivanmatmati:
An alternate solution has been merged.

PR #540, opened 2023-05-18
add TuneOptions if is missing

michal800106:
Without this set the lot of restarts occur because updated below is always true. This restarts are problematic when we want to use websockets.


oktalz:
@michal800106 thx for this, 

however can you change commit message to align with our policy?

just add prefix  `BUG/MINOR: `

so in total commit subject should be `BUG/MINOR: add TuneOptions if is missing`

PR #547, opened 2023-06-16
BUG/MINOR: quit after printing version when invoked with -v

bitfehler:
It might be a misunderstanding, but my expectation of `-v` is "print version _and exit_", which is currently not the case. Feel free to close if this is intended behavior.

### DIFF
--- a/main.go
+++ b/main.go
@@ -71,7 +71,10 @@ func main() {
 		return
 	}
 	logger.ShowFilename(false)
-	logInfo(logger, osArgs)
+	exit := logInfo(logger, osArgs)
+	if exit {
+		return
+	}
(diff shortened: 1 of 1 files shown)


ivanmatmati:
After discussion, we think it's a bug and should be amended. But we'd prefer it if you add a boolean to the function so that you can test if you exit immediately or not. Thanks.

bitfehler:
Hi @ivanmatmati, thx for the feedback. I wasn't 100% sure what you meant by "add a boolean to the function", but I figured you meant as return value for `logInfo`? I updated the patch accordingly, but please let me know if I misunderstood!

oktalz:
@bitfehler thx for this

PR #560, opened 2023-08-07
BUG/MEDIUM: support alternative healthcheck endpoint

avachette:
Wanted to add an alternative healthcheck endpoint like the PR https://github.com/haproxytech/kubernetes-ingress/pull/536 which wasn't got merged because the owner didn't reply.

I fixed the response header also.

Feel free to add remarks if something is missing.


oktalz:
@avachette thx

PR #569, opened 2023-09-02
BUG/MINOR: prevents clash on custom routes names to avoid unecessary restarts

### LINKED ISSUE #567: Order of "use_backend" lines in haproxy.cfg is inconsistent (changes every 5 seconds)
Hello,

We use the "haproxy.org/route-acl" option for selecting backends. The haproxy controller turns this into "use_backend" rules in the haproxy.cfg file. However, it seems that the order which these lines are generated is inconsistent. And this probably also causes haproxy to reload every 5 seconds (the default sync period).

It would be nice if the order was consistent.

Thanks,

Timo

fabianonunes:
Fixes #567

All custom routes are stored in the `routes.CustomRoutes` map using the backend name as the key (`ServiceNS_ServiceName_PortName`). When we encounter two distinct IngressRules, whether they are within the same Ingress or not, and they utilize different hosts or paths while pointing to the same Service, they end up sharing the same key in `CustomRoutes`. This similarity in keys results in comparison failures during every synchronization, requiring a hard reload of HAProxy.

Rather than storing the custom routes in a map keyed by backend name, this PR stores all the resulting `use_backend` directives in a slice and compares them between syncs.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-test/deep"
+
 	"github.com/haproxytech/client-native/v3/models"
 	"github.com/haproxytech/kubernetes-ingress/pkg/annotations"
 	gateway "github.com/haproxytech/kubernetes-ingress/pkg/gateways"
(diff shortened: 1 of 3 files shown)


oktalz:
@fabianonunes thx!

PR #579, opened 2023-10-07
For using latest gateway api v0.8.1

eltorio:
NamespacesFromSame and NamespacesFromSelector were promote to v1beta1 by sig


ivanmatmati:
Hi @eltorio , Did you check if there was no functional difference ? We need to check if there's more to implement.

eltorio:
@ivanmatmati according to all the changelog since 5.1 the only changes reported are
```
v1alpha2 of Gateway, GatewayClass, and HTTPRoute is no longer served
v1alpha2 of ReferenceGrant is deprecated
v1beta1 is now the storage version for ReferenceGrant
Those changes mean that:

Users and implementations that were reading or writing from v1alpha2 of
Gateway, GatewayClass, or HTTPRoute MUST upgrade to use v1beta1.
Users and implementations that were reading or writing from v1alpha2 of
ReferenceGrant SHOULD upgrade to use v1beta1.
```

PR #593, opened 2023-12-15
Upgrade S6 overlay to 3.x

PKizzle:
Changes the installation of s6-overlay to support version 3.x and alters the file structure.

Haproxy is no longer started automatically upon container start but is instead managed by kubernetes-ingress.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -40,9 +40,11 @@ FROM haproxytech/haproxy-alpine:2.8
 
 ARG TARGETPLATFORM
 
-ARG S6_OVERLAY_VERSION=2.2.0.3
+ARG S6_OVERLAY_VERSION=3.1.6.2
 ENV S6_OVERLAY_VERSION $S6_OVERLAY_VERSION
 ENV S6_READ_ONLY_ROOT=1
+ENV S6_USER=haproxy
+ENV S6_GROUP=haproxy
 
 COPY /fs /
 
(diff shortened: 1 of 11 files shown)


oktalz:
hi, @PKizzle thx for the PR

this will require a bit more testing and investigation from our side, 
primarily since we tried to upgrade to to v3 before, and it failed.
v3 does not play well with rootless environments 

this might have changed, but last time we checked that was the case.

oktalz:
also change to start haproxy from k8s is not ideal due to heatlh checks (and other probes).
its not imposible but it would change significantly how k8s sees the pod

PKizzle:
I changed it that way since haproxy would be reachable so probes might return successfully, however the configuration would still be invalid as the first config update would not have yet happened (which requires a haproxy restart / reload anyways)

dkorunic:
We have been testing S6 overlay 3.x as soon as it was released, but the main problem with adoption was that it stopped working with read-only containers. I don't think there was any progress there on the part of S6 overlay and since there was no benefit in upgrading to 3.x, we have delayed this indefinitely.

PKizzle:
If I read the [README](https://github.com/just-containers/s6-overlay/blob/master/README.md) correctly it should be possible to use s6-overlay with read-only containers by specifying `S6_READ_ONLY_ROOT=1`.

dkorunic:
I've thoroughly investigated this and sadly there are many changes that are should be reworked in this PR (namely: unnecessary changes in `pkg/haproxy/process/s6-overlay.go`, missed `haproxy` service dependancy for `ingress-controller` service, missed `SIGUSR1` handler for `s6-svscan`), especially because this also needs fixing in Helm Chart as well (ie. to make /tmp and /run auto-mounted as tmpfs on boot). Good news is that I've refactored all of that together (Helm Chart and IC) and tested both our master and `v1.10` branch and it works w/ and w/o read-only mode well, so I'll push my refactored changes and reference/credit your original PR.

dkorunic:
Required Helm Chart logic for `/tmp' and `/run` incoming in https://github.com/haproxytech/helm-charts/commit/f67966e0b066c35ba80bbdb79a07cba270934f2e.

PR #594, opened 2023-12-15
Check for successful haproxy startup

PKizzle:
Instead of assuming that haproxy API server is already reachable when starting up kubernetes-ingress wait until it is available. This fixes startup issues for systems running on low-end hardware where kubernetes-ingress would go into a crash-loop as haproxy never starts up fast enough.


oktalz:
hi @PKizzle , 

thx for the PR.
if this is the issue on low-end hardware I agree we should address it.
However, we can't remove inotify from aux file, that file is allowed to be changed over time

PKizzle:
Could you explain what you mean with the removal of inotify from the aux file? I am unable to find it in my code changes 😅

dkorunic:
This is much easier to fix with relying on S6 dependancies in S6 overlay 3.1 and with fine-tuning `startupProbe. failureThreshold` if needed. The changes here and in https://github.com/haproxytech/kubernetes-ingress/pull/593 bring too much changes for a specific case. I'll close this request and address this in incoming S6 overlay 3.1 overhaul.

PR #610, opened 2024-01-31
enable TCPRoute to use HAProxy v2 protocol to communicate with the backend

eltorio:
# For gateway API (≥ v0.5.1) #

## New annotation in TCPRoute ##
```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: TCPRoute
metadata:
  annotations:
    haproxy.org/send-proxy-v2: 'true'
```
this will enable to use HAProxy v2 protocol to communicate with the backend this is useful for transmitting the original IP of the remote client.

### DIFF
--- a/documentation/doc.yaml
+++ b/documentation/doc.yaml
@@ -1878,3 +1878,17 @@ annotations:
     version_min: "1.7"
     example:
       - "tls-alpn: http/1.1"
+  - title: send-proxy-v2
+    type: string
+    group: gateway
+    dependencies: ""
+    default: "false"
+    description:
+      - For TCPRoutes only send HAProxy v2 protocol to the backend.
+    values:
+      - anything not 'True' or 'true' will be considered false
+    applies_to:
+      - gateway
+    version_min: "1.10"
+    example:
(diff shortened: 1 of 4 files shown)


documentation/doc.yaml
```diff
@@ -1878,3 +1878,17 @@ annotations:
     version_min: "1.7"
     example:
       - "tls-alpn: http/1.1"
+  - title: send-proxy-v2
```
eltorio, line 1881: Not sure it is the good place in the documentation

pkg/gateways/gateways.go
```diff
@@ -605,6 +611,11 @@ func getBackendName(tcproute store.TCPRoute) string {
 	return tcproute.Namespace + "_" + tcproute.Name
 }
 
+// getBackendUseProxyV2 provides backend send-proxy-v2 from backend attributes.
+func getBackendUseProxyV2(tcproute store.TCPRoute) bool {
+	return tcproute.BackendRefs[0].ProxyV2
```
eltorio, line 616: With Gateway API validation backendRefs[0] always exists

PR #627, opened 2024-03-09
BUG/MINOR: prevents unconditional reloads when prometheus is enabled and controller namespace isn't `haproxy-controller`

fabianonunes:
When the `--prometheus` flag is passed, the Prometheus backend check always fails if the controller namespace isn't `haproxy-controller`. Consequently, all events received by the controller trigger a reload of HAProxy.

```text
2024/03/09 12:32:14 INFO    instance/configuration.go:21 [transactionID=8a5a139f-1b65-4faf-8e40-dc467a54dcfb] reload required : creation/modification of prometheus endpoint
```

Since the backend name depends on the namespace, it shouldn't be hardcoded.

### DIFF
--- a/pkg/handler/prometheus.go
+++ b/pkg/handler/prometheus.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


oktalz:
@fabianonunes thx!

Issue #643, opened 2024-04-15
Deployment yaml still point to privileged port after s6 upgrade


dosmanak:
After upgrade on to v1.11.2 we noticed the haproxy is not reachable from outside. So we found there has been upgrade of s6 which moved to nonroot container which need non-privileged ports.
The [commit](https://github.com/haproxytech/kubernetes-ingress/commit/e54beaee2260b0531d0749a46e8ab2455a009f7b) ommits the change in deploy manifest.

Also the default setting of `ssl-redirect-port` ceases to be very handful since it now defaults to non-privileged port which is rare on load balancers.

I suggest to change only [targetPort](https://github.com/haproxytech/kubernetes-ingress/blob/e54beaee2260b0531d0749a46e8ab2455a009f7b/deploy/haproxy-ingress-daemonset.yaml#L212) of http and https backends in haproxy service to maintain it as much backwards compatible as possible.

ivanmatmati:
Hi @dosmanak , thanks for reporting, we'll check that.

dosmanak:
The PR has been merged.

PR #644, opened 2024-04-16
BUG/MAJOR: deploy: respect change of default port to 8080 and 8443

dosmanak:
this fixes the commit e54beae

issue:  #643

### DIFF
--- a/deploy/haproxy-ingress-daemonset.yaml
+++ b/deploy/haproxy-ingress-daemonset.yaml
@@ -127,6 +127,7 @@ metadata:
   name: haproxy-kubernetes-ingress
   namespace: haproxy-controller
 data:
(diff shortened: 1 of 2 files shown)


dosmanak:
It seems we are the only one who deploys using
https://github.com/haproxytech/kubernetes-ingress/blob/master/deploy/haproxy-ingress-daemonset.yaml

remram44:
This really deserves an errata in the release note or something. It takes down people's entire cluster of applications.

remram44:
Another important detail for people running into this: you will need to set `ssl-redirect-port: "443"` in the configmap, otherwise all ingresses hit over plain HTTP will redirect to `https://...:8443/` and fail.

I am considering other ingress controllers. This update was bullshit.

PR #657, opened 2024-06-14
BUG: update IC pod-serving port to match default

rwaweber:
mismatched from service object

Hey all - noticed when updating our deployment that nothing is being served on either port 443 or port 80 on the pod anymore. Figured I'd correct this to the current defaults.

Thanks!


oktalz:
hi @rwaweber 

thx.

PR #661, opened 2024-07-09
supporting thread pin on http https and healthz frontend

cxfcxf:
We are currently using ingress controller as external mode, so it would be nice that we can control the thread pin on http/https/healthz for better performance.

The PR is adding those flag to support thread pin for ingress controller to set them on haproxy config.

i don't think its a necessary requirement in term of if the controller is running inside Kubernetes cluster, but it would be a great feature to provide for people running it as external mode.

### DIFF
--- a/pkg/utils/flags.go
+++ b/pkg/utils/flags.go
@@ -92,9 +92,12 @@ type OSArgs struct {
 	ControllerPort             int            `long:"controller-port" description:"port to listen on for controller data: prometheus, pprof" default:"6060"`
 	HTTPBindPort               int64          `long:"http-bind-port" default:"8080" description:"port to listen on for HTTP traffic"`
 	HTTPSBindPort              int64          `long:"https-bind-port" default:"8443" description:"port to listen on for HTTPS traffic"`
+	HTTPBindThread             string         `long:"http-bind-thread" description:"default http service bind thread params eg: 1-1" default:""`
+	HTTPSBindThread            string         `long:"https-bind-thread" description:"default https service bind thread params eg: 1-1" default:""`
 	SyncPeriod                 time.Duration  `long:"sync-period" default:"5s" description:"Sets the period at which the controller syncs HAProxy configuration file"`
 	CacheResyncPeriod          time.Duration  `long:"cache-resync-period" default:"10m" description:"Sets the underlying Shared Informer resync period: resyncing controller with informers cache"`
 	HealthzBindPort            int64          `long:"healthz-bind-port" default:"1042" description:"port to listen on for probes"`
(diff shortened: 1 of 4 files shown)


oktalz:
hi @cxfcxf 
thx for the PR

we will look at it.

in the mean time CI is complaining about commit message. in short 
```
support thread pin on frontend/status
```
should in this case be
```
MINOR: support thread pin on frontend/status
```

pkg/utils/flags.go
```diff
@@ -92,9 +92,12 @@ type OSArgs struct {
...
 	HTTPBindPort               int64          `long:"http-bind-port" default:"8080" description:"port to listen on for HTTP traffic"`
 	HTTPSBindPort              int64          `long:"https-bind-port" default:"8443" description:"port to listen on for HTTPS traffic"`
+	HTTPBindThread             string         `long:"http-bind-thread" description:"default http service bind thread params eg: 1-1" default:""`
+	HTTPSBindThread            string         `long:"https-bind-thread" description:"default https service bind thread params eg: 1-1" default:""`
 	SyncPeriod                 time.Duration  `long:"sync-period" default:"5s" description:"Sets the period at which the controller syncs HAProxy configuration file"`
 	CacheResyncPeriod          time.Duration  `long:"cache-resync-period" default:"10m" description:"Sets the underlying Shared Informer resync period: resyncing controller with informers cache"`
 	HealthzBindPort            int64          `long:"healthz-bind-port" default:"1042" description:"port to listen on for probes"`
+	HealthzBindThread          string         `long:"healthz-bind-thread" description:"default healthz service bind thread params eg: 1-1" default:"1-1"`
```
oktalz, line 100: for others default is `""` so its does not change behavior (I still need to check that), but this one has default that is not "".
Any reason why this should be set ? it is a breaking change that in general we try to avoid when possible

cxfcxf:
> hi @cxfcxf thx for the PR
> 
> we will look at it.
> 
> in the mean time CI is complaining about commit message. in short
> 
> ```
> support thread pin on frontend/status
> ```
> 
> should in this case be
> 
> ```
> MINOR: support thread pin on frontend/status
> ```

i did a reword on commit, so it should be fixed.

oktalz:
@cxfcxf 

> i did a reword on commit, so it should be fixed.
we now have two commits with same subject, and they should be squashed.

its best that you do that, however I can do that too (if no objections)

cxfcxf:
> @cxfcxf
> 
> > i did a reword on commit, so it should be fixed.
> > we now have two commits with same subject, and they should be squashed.
> 
> its best that you do that, however I can do that too (if no objections)

ok i squash second commit.

oktalz:
@cxfcxf thx!, 

merged

PR #667, opened 2024-08-07
MINOR: support thread pin on stats

cxfcxf:
an addition to https://github.com/haproxytech/kubernetes-ingress/pull/661

i think i missed stats binding thread on that PR. this adds it


oktalz:
@cxfcxf thx.

PR #692, opened 2025-01-09
MINOR: Refine route-acl rules to prevent unintended prefix matches

fabianonunes:
Since `route-acl` annotated rules take precedence over others, this PR updates its behavior to ensure they do not unintentionally overwrite other rules that share the same word prefix.

For example, a rule matching the path prefix `/api` should not inadvertently handle requests to `/apiary`.

To address this, the rule `{ path -m beg /api }` has been replaced with a alternative that validates the URL's termination, ensuring it matches only `/api$` or `/api/.*`:

```
# before this PR:
use_backend app_api_http if { var(txn.host) -m str api.demo } { path -m beg /api } { ... }

# after:
use_backend app_api_http if { var(txn.host) -m str api.demo } { path -m reg ^/api($|/) } { ... }
```

For better maintainability, this PR also refactors the `AddCustomRoute` function to eliminate redundancy introduced in commit c28d620aee948a8e912c9a52b4be4ad4d8cd6c82. The updated code removes repetition without add extra spaces.

### DIFF
--- a/.aspell.yml
+++ b/.aspell.yml
@@ -1,6 +1,7 @@
 mode: commit
 min_length: 3
 allowed:
+  - acl
   - aspell
   - repo
   - yaml
--- a/pkg/route/route.go
+++ b/pkg/route/route.go
@@ -108,17 +108,9 @@ func AddCustomRoute(route Route, routeACLAnn string, api api.HAProxyClient) (err
 	}
 	if route.Path.Path != "" {
 		if route.Path.PathTypeMatch == store.PATH_TYPE_EXACT {
-			if routeCond == "" {
-				routeCond = fmt.Sprintf("{ path %s }", route.Path.Path)
-			} else {
-				routeCond = fmt.Sprintf("%s { path %s }", routeCond, route.Path.Path)
-			}
+			routeCond = fmt.Sprintf("%s{ path %s }", routeCond, route.Path.Path)
 		} else {
-			if routeCond == "" {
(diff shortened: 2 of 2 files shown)


fabianonunes:
I'd like to know if there's anything I can do to facilitate the merge of this PR.

I understand that using regex might have some performance impact (though I'm not certain about this). If that's a concern, we could implement two separate rules instead:

```haproxy
# matches all paths that begin with `/api/`
use_backend app_api_http if { var(txn.host) -m str api.demo } { path -m beg /api/ } { ... }
# matches all requests with the exact path `/api`
use_backend app_api_http if { var(txn.host) -m str api.demo } { path -m str /api } { ... }
```

This alternative approach would achieve the same goal of preventing unintended matches (like /apiary) while potentially avoiding any regex performance overhead, if that's the case. Please let me know your thoughts or if there are any other concerns I can address to help move this PR forward.

Regarding the failed check `HAProxy check commit message`, I've already added the word `acl` to the dictionary and force-pushed, but the job hasn't been re-executed yet.

fabianonunes:
bump

ivanmatmati:
Hi @fabianonunes , we need to discuss if this is something we want to integrate because the documentation says:

> Note that this annotation is not compatible with an Ingress having multiple paths that will match a request. Without this annotation, the precedence is given first to the longest matching path. But with the annotation, the first use_backend rule in the config that matches the request will be used.

Your PR could be an enhancement but we need to check if we want to go this direction. There's also a failing e2e test, can you check what happened ?

PR #701, opened 2025-02-26
MINOR: refactor quic configuration

PKizzle:
Splits up the single update function into smaller helper functions and reduces code duplication.
Furthermore, it checks whether ssl-offload is correctly enabled.


PKizzle:
Oups, it seems like aspell thinks that "quic" is a typo 😄

PKizzle:
@oktalz Is there anything I can do from my side?

PR #704, opened 2025-03-05
BUG: Make ingress pathType "Exact" always override pathType "Prefix"

### LINKED ISSUE #703: Ingress pathType "Prefix" can override pathType "Exact"
Let's assume we have a web service with prefix `/foo`. There are many sub-paths for `/foo/...`. Let's also assume we want to have different timeouts for `/foo` and `/foo/.*`.

We would expect the following to work:

```yaml
kind: Ingress
metadata:
  name: subpath-foo
  annotations:
    haproxy.org/standalone-backend: "true"
    haproxy.org/timeout-server: 5s
spec:
  ingressClassName: haproxy
  rules:
  - host: example.tld 
    http:
      paths:
      - path: /foo
        pathType: Prefix
        backend:
          service:
            name: foo-service
            port:
              name: http
---
kind: Ingress
metadata:
  name: just-foo
  annotations:
    haproxy.org/standalone-backend: "true"
    haproxy.org/timeout-server: 10s
spec:
  ingressClassName: haproxy
  rules:
  - host: example.tld 
    http:
      paths:
      - path: /foo
        pathType: Exact
        backend:
          service:
            name: foo-service
            port:
              name: http
```

The annotation `haproxy.org/standalone-backend`ensures two separate backends are created so that the different timeouts do not fight each other.
Now we have the problem that `/etc/haproxy/maps/path-exact.map` has the following contents:

```
example.tld/foo          ing_default_subpath-foo_http.dbe4b28f9c4404fc36a7ae76ead9a37d
example.tld/foo          ing_default_just-foo_http.dbe4b28f9c4404fc36a7ae76ead9a37d
```

That means the actual exact match from ingress `just-foo` will never match since pathType `Prefix` patterns are also put into this pattern file.
[Here](https://github.com/haproxytech/kubernetes-ingress/blob/04d271ccc641fb9295b0ea1c7cf7af5f9397e52d/pkg/route/route.go#L90) we can find the responsible code:

```golang
	path := route.Path.Path
	switch {
	case route.Path.PathTypeMatch == store.PATH_TYPE_EXACT:
		mapFiles.MapAppend(PATH_EXACT, route.Host+path+"\t\t\t"+value)
	case path == "" || path == "/":
		mapFiles.MapAppend(PATH_PREFIX, route.Host+"/"+"\t\t\t"+value)
	case route.Path.PathTypeMatch == store.PATH_TYPE_PREFIX:
		path = strings.TrimSuffix(path, "/")
		mapFiles.MapAppend(PATH_EXACT, route.Host+path+"\t\t\t"+value)
		mapFiles.MapAppend(PATH_PREFIX, route.Host+path+"/"+"\t\t\t"+value)
	case route.Path.PathTypeMatch == store.PATH_TYPE_IMPLEMENTATION_SPECIFIC:
		path = strings.TrimSuffix(path, "/")
		mapFiles.MapAppend(PATH_EXACT, route.Host+path+"\t\t\t"+value)
		mapFiles.MapAppend(PATH_PREFIX, route.Host+path+"\t\t\t"+value)
	default:
		return fmt.Errorf("unknown path type '%s' with backend '%s'", route.Path.PathTypeMatch, route.BackendName)
	}
```

I do understand why `Prefix` paths must have an entry in `path-exact.map`: The [definition](https://kubernetes.io/docs/concepts/services-networking/ingress/#examples) of ingress resources mandates that `/foo` must match `/foo` and `/foo/bar` but never `/foobar`. Therefore we add `/foo` to exact matches and `/foo/` to prefix matches (via `map_beg`).

The only way I see to resolve this issue in a clean and backward-compatible way is to introduce a new map file which will contain all exact matches from pathType `Prefix` only and it will be matched if there was no match in `path-exact.map` which will then contain ONLY pathType `Exact` patterns.

I will follow up with a PR.

phihos:
Fixes #703.

We introduce a new map `path-prefix-exact` which contains exact match patterns for ingress pathType `Prefix`. From now on the map `path-exact` will only contain patterns from pathType `Exact`. The generated HAProxy config will check `path-exact` first,  `path-prefix-exact` second and `path-prefix` third. That way pathType `Exact` always wins.

### DIFF
--- a/deploy/tests/e2e/client.go
+++ b/deploy/tests/e2e/client.go
@@ -202,7 +202,7 @@ func runtimeCommand(command string) (result []byte, err error) {
 	if err != nil {
 		return
 	}
-	result = make([]byte, 1024)
+	result = make([]byte, 2048)
 	_, err = conn.Read(result)
 	conn.Close()
 	return
@@ -217,7 +217,7 @@ func GetHAProxyMapCount(mapName string) (count int, err error) {
 	scanner := bufio.NewScanner(bytes.NewReader(result))
 	for scanner.Scan() {
 		line := scanner.Text()
-		if strings.Contains(line, mapName) {
+		if strings.Contains(line, mapName+".map") {
 			r := regexp.MustCompile("entry_cnt=[0-9]*")
 			match := r.FindString(line)
 			nbr := strings.Split(match, "=")[1]
--- a/deploy/tests/e2e/map-updates/update_test.go
+++ b/deploy/tests/e2e/map-updates/update_test.go
@@ -30,17 +30,27 @@ func (suite *MapUpdateSuite) Test_Update() {
 			suite.tmplData.Paths = append(suite.tmplData.Paths, strconv.Itoa(i))
 		}
 		oldInfo, err := e2e.GetGlobalHAProxyInfo()
-		oldCount, err := e2e.GetHAProxyMapCount("path-prefix")
+		oldCountExact, err := e2e.GetHAProxyMapCount("path-exact")
+		suite.Require().NoError(err)
+		oldCountPrefixExact, err := e2e.GetHAProxyMapCount("path-prefix-exact")
+		suite.Require().NoError(err)
+		oldCountPrefix, err := e2e.GetHAProxyMapCount("path-prefix")
 		suite.Require().NoError(err)
 		suite.Require().NoError(suite.test.Apply("config/ingress.yaml.tmpl", suite.test.GetNS(), suite.tmplData))
 		suite.Require().Eventually(func() bool {
 			newInfo, err := e2e.GetGlobalHAProxyInfo()
 			suite.Require().NoError(err)
-			count, err := e2e.GetHAProxyMapCount("path-prefix")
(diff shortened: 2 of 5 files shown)


phihos:
I fixed the broken e2e test and the commit message. Can somebody trigger the tests again?

phihos:
The test `TestMapUpdateSuite/Test_Update/Update` now checks all three `path-*` map files.

phihos:
The e2e test  failure is odd. I ran `./deploy/tests/create.sh` and then `make e2e` without any failure on my machine. What am I missing here?

oktalz:
@phihos I'll check, it wouldn't be the first time CI is misbehaving

phihos:
Commenting to remove stale label. And updates on the CI issues?

phihos:
Comment to remove stale label.

ivanmatmati:
Hi @phihos , can you check why there's an e2e test failure ?

phihos:
Hi @ivanmatmati, the tests still run fine locally. I rebased them on latest master. Can you trigger them to see if the issue is still there?

PR #731, opened 2025-08-08
MEDIUM: config: add disable-ingress-status-update configuration flag

vriabyk:
This PR introduces a new controller configuration flag. When enabled, this flag prevents the controller from updating the status field of Ingress resources. By default, status updates remain enabled.

New flag:
```
--disable-ingress-status-update (boolean, default: false)
```

When set, the controller will skip updating the LoadBalancer status field in managed Ingress resources.
Useful for scenarios where another controller or external process manages Ingress status, or when status updates are not desired. It won't set status field and it will be just empty:

```
$ k get ingress test-ingress-99999 -o jsonpath='{.status}' | jq . -r
{
  "loadBalancer": {}
}
```

Code changes:

* Added the DisableIngressStatusUpdate field to the configuration struct.
* Passed the flag through the controller and status update logic.
* Updated the UpdateStatus method to respect the flag.
* Updated documentation

The flag is now listed in the arguments table and has a dedicated section describing its usage and example. If the flag is not specified, status updates are enabled by default.

### DIFF
--- a/documentation/controller.md
+++ b/documentation/controller.md
@@ -1,3 +1,4 @@
+
 # ![HAProxy](../assets/images/haproxy-weblogo-210x49.png "HAProxy")
 
 ## HAProxy kubernetes ingress controller
--- a/documentation/controller.md
+++ b/documentation/controller.md
@@ -1,4 +1,3 @@
-
 # ![HAProxy](../assets/images/haproxy-weblogo-210x49.png "HAProxy")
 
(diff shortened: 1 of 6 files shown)


Sea-you:
This is certainly a feature, I'm interested in!

vriabyk:
Hi @oktalz , second option works fine for me :) please do it, thank you!

Issue #734, opened 2025-09-04
Subdomain wildcard does not work together with route-acl service annotation


SF97:
**APP VERSION**: 3.1.11
**CHART**: kubernetes-ingress-1.44.6

It's possible to use [route-acl](https://www.haproxy.com/documentation/kubernetes-ingress/community/configuration-reference/service/#route-acl) service annotation to specify paths against regular expressions. For instance, if I want the service to only match paths with `path-in-bug-repro`, I can add the Service annotation `haproxy.org/route-acl: path_reg path-in-bug-repro$`, using the following helm value file

```YAML
# helm upgrade --install -f ./values.yaml haproxy-bug oci://registry-1.docker.io/bitnamicharts/nginx
ingress:
  enabled: true
  hostname: "example.local"
  ingressClassName: "haproxy"
  tls: false

service:
  annotations:
    haproxy.org/route-acl: path_reg path-in-bug-repro$
  type: ClusterIP

# Define a custom server block.
# The content is a multi-line string containing the Nginx configuration.
serverBlock: |
  server {
    listen 0.0.0.0:8080;
    server_name _;

    location / {
      default_type text/html;

      <!DOCTYPE html>
      <html lang="en">
      <body>
          <h1>BUG REPRO</h1>
      </body>
      </html>
      ';
    }
  }
```

And requests to said path will match:

<img width="1030" height="333" alt="Image" src="https://github.com/user-attachments/assets/16f89250-6448-4ff7-88eb-a669914239c5" />

Other paths will not:

<img width="1437" height="144" alt="Image" src="https://github.com/user-attachments/assets/fcfc47d7-8741-4749-8324-f515230df7cd" />

However, when trying to use this feature in combination with wildcard  subdomains, it does not work. Using the YAML (notice the wildcard in the hostname)

```yaml
# helm upgrade --install -f ./values.yaml haproxy-bug oci://registry-1.docker.io/bitnamicharts/nginx
ingress:
  enabled: true
  hostname: "*.example.local"
  ingressClassName: "haproxy"
  tls: false

service:
  annotations:
    haproxy.org/route-acl: path_reg path-in-bug-repro$
  type: ClusterIP

# Define a custom server block.
serverBlock: |
  server {
    listen 0.0.0.0:8080;
    server_name _;

    location / {
      # Set the Content-Type header so browsers render the HTML correctly.
      default_type text/html;

      <!DOCTYPE html>
      <html lang="en">
      <body>
[...]

PR #735, opened 2025-09-18
BUG/MINOR: controller: Fix wildcard host matching with route-acl

SF97:
When an Ingress resource used a wildcard host (e.g., `*.example.com`) in combination with the `haproxy.org/route-acl` service annotation, the controller would generate an incorrect HAProxy ACL. It used an exact string match (`-m str`) on the literal value `*.example.com`, which would fail to match any intended subdomains.

This patch modifies the route generation logic to inspect the hostname. If the host begins with an asterisk ('*'), it now correctly generates an ACL using a suffix match (`-m end`) and removes the leading asterisk from the hostname string.

For non-wildcard hosts, the original behavior of using an exact string match (`-m str`) is preserved.

Fixes: #734

### DIFF
--- a/.aspell.yml
+++ b/.aspell.yml
@@ -57,3 +57,5 @@ allowed:
   - frontent
   - pprof
   - preload
+  - hostname
+  - str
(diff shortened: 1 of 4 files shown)


SF97:
Would love to have this reviewed @oktalz :) 

This would help us a lot, it's blocking a release on our side, and it benefits all HAProxy users with the same use case

oktalz:
@SF97 I'm sorry for late response I'm merging this.
thx for the PR!

Issue #737, opened 2025-10-15
First port number is used for all endpoints even if there are multiple different port numbers


teqwve:
Hi! It looks like the ingress uses the first port number encounteted for given Service port for all addresses. It may result in incorrect haproxy configuration if there are multiple endpoint slices with different port numbers and the same port name - for some servers incorrect port is used. Such configuration can be created by e.g. targeting Pods from multiple Deployments by one Service (I attached an example below). Such configuration is handled by k8s components so I think this may be regarded as a bug in ingress?

It looks like the information about different ports is lost in [`getEndpoints`](https://github.com/haproxytech/kubernetes-ingress/blob/8d87cfcabbba305293ae3a8ddb60634274d84c33/pkg/store/events.go#L123) where only the `Port` from the first `portEndpoints` collected for given Service port is copied to `PortEndpoints` together with all addresses. 

I briefly looked at the code and it looks like fix could be to change `getEndpoints` to return a list of 'tuples' (address, port), change [`RuntimeBackend.Endpoints`](https://github.com/haproxytech/kubernetes-ingress/blob/8d87cfcabbba305293ae3a8ddb60634274d84c33/pkg/store/types.go#L81) to keep this list and then refactor all places to handle different ports. Unfortunately it looks like [`SyncBackendSrvs`](https://github.com/haproxytech/kubernetes-ingress/blob/8d87cfcabbba305293ae3a8ddb60634274d84c33/pkg/haproxy/api/runtime.go#L163) would have to be basically rewritten (it's where I gave up at the moment, I can try to prepare a PR if you like the idea) :/

WDYT?

---

With following manifests:

```
---
apiVersion: v1
kind: Pod
metadata:
  name: test-1
  labels:
    app: test
spec:
  containers:
    - name: app
      image: python:3.12
      command:
        - python3
        - -m
        - http.server
        - "8081"
      ports:
        - containerPort: 8081
          name: http
---
apiVersion: v1
kind: Pod
metadata:
  name: test-2
  labels:
    app: test
spec:
  containers:
    - name: app
      image: python:3.12
      command:
        - python3
        - -m
        - http.server
        - "8082"
      ports:
        - containerPort: 8082
          name: http
---
apiVersion: v1
kind: Service
metadata:
  name: test
spec:
  type: ClusterIP
  ports:
    - name: http
      port: 80
      targetPort: http
  selector:
    app: test
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
[...]

PR #777, opened 2026-02-08
BUG/MEDIUM: handle different endpoint ports

teqwve:
This pull request is a fix for https://github.com/haproxytech/kubernetes-ingress/issues/737

With this change ingress maintains a separate port for each ip address:
- in new RuntimeEndpoint structure for new endpoints
- in HAProxySrv.Port for endpoints that have been configured on backend

and uses per-address port for configuration (from HAProxySrv.Port) in places where common RuntimeBackend.Endpoint.Ports was used before.

Previously k8s representation (PortEndpoints) was used for passing information about new endpoints and only one (the first) port number was used for all addresses even if there were different ports. The new structure does not loose this information.

Apart from automatic tests I briefly tested it manually with ingress / endpoints and very briefly with tcp crd. I haven't manually tested other crds and tcp route. Please let me know if there are other things worth testing I can add new e2e tests or test them manually.

### DIFF
--- a/deploy/tests/e2e/different-ports/config/deploy.yaml.tmpl
+++ b/deploy/tests/e2e/different-ports/config/deploy.yaml.tmpl
@@ -1,0 +1,44 @@
+{{ range $instanceIndex, $instance := .Instances }}
+---
+kind: Deployment
+apiVersion: apps/v1
+metadata:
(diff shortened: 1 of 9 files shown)


oktalz:
@teqwve thx

PR #788, opened 2026-02-24
MINOR: Revise SSL Redirection on default HTTPS port

### LINKED ISSUE #642: http -> https default redirect not working by default
I'm using the latest kubernetes-ingress.
If a http request arrives at :80 to a valid ingress host it redirects to the ingress host:8443.
The 8443 comes from the containerport where the kubernetes-ingress listening. It sould redirect to https://ingresshost without any :port.

At the docs I see that there is a ssl-redirect-port option. It works but i got https://ingresshost:443 if I set this option to 443.
There is a config option for the redirect without port (since the default is 443 at the browsers)?

BuJo:
* The default for the controller is to generate redirects to port `8443`, as this port can be bound to in a rootless container.
* Running the container in a Kubernetes setting having a service port mapping to port 443 introduces the need to change this default configuration.
* Having the default HTTPs port `443` appended to the redirect breaks certain caching behaviours and is thus not ideal.
* This commit aims to purely switch the scheme from HTTP to HTTPs without changing anything else about the request, the only user visible change being that the `:443` in the redirect URL is no longer visible.
* fixes #642

### DIFF
--- a/.aspell.yml
+++ b/.aspell.yml
@@ -72,5 +72,7 @@ allowed:
   - userlist
   - tmp
   - kubectl
+  - Kubernetes
+  - HTTPS
   - PEM
   - redact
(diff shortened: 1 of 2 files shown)


tim-van-dijkhuizen:
Bump! I've experienced this exact same issue and implemented a similar solution. Was about to create a PR for it, but someone (obviously) beat me to it. Would really like to see this merged :)

PR #792, opened 2026-03-06
BUG/MINOR: Prevent user addition with whitespace in password hash

markkrj:
If password hash contains whitespace of any kind, haproxy fails to validate the config and will not start or reload.

According to `man 5 crypt`:
```
Hashed passphrases are always entirely printable ASCII, and do not contain any whitespace or the characters
     ‘:’, ‘;’, ‘*’, ‘!’, or ‘\’.
```
I don't know if it must be this controller's job to validate all of that. But, whitespace in the password does brake it from generating valid HAProxy config, so that is a bug.
Also, I think that emitting a warning is helpful for admins to figure out why some user cannot login.

/cc @ivanmatmati: I talked to you on Slack
/cc @hdurand0710: You fixed a bug related

### DIFF
--- a/pkg/annotations/ingress/basicAuth.go
+++ b/pkg/annotations/ingress/basicAuth.go
@@ -1,8 +1,10 @@
 package ingress
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
+	"unicode"
 
 	"github.com/haproxytech/kubernetes-ingress/pkg/annotations/common"
 	"github.com/haproxytech/kubernetes-ingress/pkg/haproxy/rules"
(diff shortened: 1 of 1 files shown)


pkg/annotations/ingress/basicAuth.go
```diff
@@ -76,15 +78,14 @@ func (a ReqAuthAnn) Process(k store.K8s, annotations ...map[string]string) (err
...
+			// If the pwd is empty or has spaces, we do not create the user in the userlist
 			// This is not a valid setting
-			if len(pwd) > 0 {
-				if pwd[len(pwd)-1] == '\n' {
-					// logger.Warningf("Ingress %s/%s: basic-auth: password for user %s ends with '\\n'. Ignoring last character.", ingress.Namespace, ingress.Name, u)
-					pwd = pwd[:len(pwd)-1]
-				}
-				a.parent.authRule.Credentials[u] = pwd
+			if len(pwd) < 1 || bytes.ContainsFunc(pwd, unicode.IsSpace) {
+				logger.Warningf("Ingress %s/%s: basic-auth: password for user %s is empty or has spaces. Ignoring it.", a.parent.ingress.Namespace, a.parent.ingress.Name, u)
```
oktalz: when basic-auth is configured through the main ConfigMap, this will panic,
a.parent.ingress can be nil (there is similar bug potentially unrelated to this, but I will cover it)

warning here also will flood the logs (I think, need to check), but i can solve that later.

oktalz (changes requested):

oktalz:
hi @markkrj 
thx for this, it founds some things we need to change additionally, I think different warning message would be ok here and I can deal with rest of things I found

markkrj:
@oktalz What do you think of this approach?
In an incorrect config, the log will be flooded, but also, it took me some time to find the offender when my IC crashed.
This way, the admin would immediately tell which ingress/secret is wrong.

PR #793, opened 2026-03-10
BUILD/MINOR: Switch order on inside c.Start(), so restart will work and run Start() synchronously and split SyncData() into its own goroutine

cxfcxf:
**Problem**

When running the controller in external mode (e.g. under systemd), systemctl restart hangs ~50% of the time. The controller gets stuck and never processes Kubernetes events.

**Root Cause**

On restart, the old HAProxy may still be alive briefly. Start() previously ran APICommitTransaction calls before Service("start"), which stripped the daemon directive from the on-disk config. When the old HAProxy finally exited and a new one launched, it ran in the foreground, blocking forever.

**Fix**

- Move Service("start") to the top of Start(), before any API transactions can rewrite the config. This restores the ordering from the 3.0.x series.
- Extract SyncData() out of Start() into its own goroutine and run Start() synchronously before launching the event producer. This ensures full initialization completes before any events or signals can arrive.

### DIFF
--- a/deploy/tests/integration/base-suite.go
+++ b/deploy/tests/integration/base-suite.go
@@ -152,7 +152,8 @@ func (suite *BaseSuite) StartController() {
 	annotations.InitCfgSnippet()
 	annotations.DisableConfigSnippets(testController.OSArgs.DisableConfigSnippets)
 
-	go testController.Controller.Start()
+	testController.Controller.Start()
+	go testController.Controller.SyncData()
 }
 
 func (suite *BaseSuite) StopController() {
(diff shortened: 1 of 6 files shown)


oktalz:
hi @cxfcxf 
is there a change you rebase this PR ? 
it has some conflicts with current HEAD on master

otherwise I can merge it, but not through github interface

cxfcxf:
> hi @cxfcxf is there a change you rebase this PR ? it has some conflicts with current HEAD on master
> 
> otherwise I can merge it, but not through github interface

done

oktalz:
@cxfcxf 
I need to go into more details here, 
while it looks OK in general, it might have some unexpected consequences. 
I'm still trying to figure out, if we are safe with change

PR #798, opened 2026-03-24
Case-insensitive path matching support added for ingress path

mustafasemihbay:
Normally, haproxy-kubernetes-ingress performs path matching in a case-sensitive manner. However, ingress-nginx supports case-insensitive path matching. During migrations from ingress-nginx to haproxy-kubernetes-ingress, some projects may require case-insensitive path matching behavior.

This change introduces an optional feature that enables case-insensitive path matching.

The feature can be enabled by passing an extra argument in the Helm chart configuration:

extraArgs: 
 - --case-insensitive-path-match
 
 
 kubernetes ingress example;
 
 spec:
  ingressClassName: haproxy
  rules:
    - host: example.com
      http:
        paths:
          - path: /example-api
            pathType: ImplementationSpecific
            backend:
              service:
                name: example-api
                port:
                  number: 80
                  

this ingress path match with example.com/example-api, example.com/Example-api, example.com/EXAMPLE-api etc.

### DIFF
--- a/documentation/doc.yaml
+++ b/documentation/doc.yaml
@@ -438,6 +438,14 @@ image_arguments:
     version_min: "3.2"
     values:
       - Boolean value, just need to declare the flag
+  - argument: --case-insensitive-path-match
+    description: |-
+        Enable case-insensitive path matching for Ingress resources.
+        By default, path matching is case-sensitive. If enabled, HAProxy will perform case-insensitive matching for all Ingress paths. 
+    example: --case-insensitive-path-match
+    version_min: "3.2"
+    values:
+      - Boolean value, just need to declare the flag to enable
 groups:
   config-snippet:
     header: |-
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -288,6 +288,12 @@ func (c *HAProxyController) setupHAProxyRules() error {
 			ForwardedProto: true,
 		}, false),
 	)
+
+	pathExpr := "path"
+	if c.osArgs.CaseInsensitive {
+		pathExpr = "path,lower"
+	}
+
 	for _, frontend := range []string{c.haproxy.FrontHTTP, c.haproxy.FrontHTTPS} {
 		errs.Add(
 			// txn.base var used for logging
@@ -300,7 +306,7 @@ func (c *HAProxyController) setupHAProxyRules() error {
 			c.haproxy.AddRule(frontend, rules.ReqSetVar{
 				Name:       "path",
 				Scope:      "txn",
-				Expression: "path",
+				Expression: pathExpr,
 			}, false),
 			c.haproxy.AddRule(frontend, rules.ReqSetVar{
 				Name:       "host",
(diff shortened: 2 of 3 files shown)


ivanmatmati:
Hi @mustafasemihbay , thanks for your submission. Can you make the required changes as reported by the above checks ? Have a great day.

ivanmatmati:
Hi @mustafasemihbay , can you check your expected behavior with case insensitivity matching in this setting:
```yaml
spec:
ingressClassName: haproxy
rules:
- host: example.com
http:
paths:
- path: /Example-api
pathType: ImplementationSpecific
backend:
service:
name: example-api
port:
number: 80
```
Notice the capital letter for the path and then try to match it.

maharshi9c:
Is there any update on this issue 
## Description

HAProxy Kubernetes Ingress Controller (v3.2.5, Helm chart 1.48.0) performs **case-sensitive** path matching with no working workaround available in the current release.

### The problem

Ingress paths stored in `path-prefix.map` and `path-exact.map` are matched case-sensitively against incoming request paths. When a client sends a path that differs only in case from the defined ingress path, HAProxy returns **HTTP 404** even though the path logically matches.

**Example:**
- Ingress defines: `/api/ForgotUsernamePassword/validate-credentials`
- Client sends:   `/api/forgotUsernamePassword/validate-credentials`
- Result: **HTTP 404** — no match in the map

This broke a production service after migration from Nginx (which matches paths case-insensitively by default).

---

## Expected Behaviour

One of the following should work but its not working - getting same 404 issue 

1. **Inline `(?i)` regex flag** in the ingress path should enable case-insensitive matching for that path, e.g.:
   ```yaml
   paths:
     - path: /(?i)api/ForgotUsernamePassword/.*
       pathType: ImplementationSpecific

maharshi9c:
Steps to Reproduce
1. Create an ingress with haproxy.org/use-regex: "true" and a path using (?i)

kind: Ingress
metadata:
  name: test-case-insensitive
  namespace: hqy-applications
  annotations:
    haproxy.org/ingress.class: "haproxy"
    haproxy.org/use-regex: "true"
    haproxy.org/path-match: "regex"
spec:
  ingressClassName: haproxy
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /(?i)api/MyService/.*
            pathType: ImplementationSpecific
            backend:
              service:
                name: my-service
                port:
                  number: 80
                  
    Note on Kubernetes validation: The path (?i)/api/... (flag before the /) is rejected by Kubernetes with: "must be an absolute path". The only form that passes validation is /(?i)api/... (flag after the leading /), which is what we used.

dkorunic:
Let's not keep this stale, we also have a PR on Helm Chart side as well.

dkorunic:
Revive.

oktalz:
> Hi @mustafasemihbay , thanks for your submission. Can you make the required changes as reported by the above checks ? Have a great day.

@mustafasemihbay can you run `make doc` and then amend the changes, also 
can you change the commit message from `case-insensitive path matching support added` to `MINOR: case-insensitive path matching support added` ?

please note that we have a rather strict policy around commit messages too, but that change should be enough.

also @ivanmatmati mentioned some concerns around capital letter for the path that needs to be addressed

Issue #795, opened 2026-03-16
HAProxy runtime backends drift between ingress replicas after application deployment


asjonos:
**Description**
We are seeing inconsistent live backend state between HAProxy Kubernetes Ingress replicas after redeploying an application behind a Service.

The generated haproxy.cfg on disk is updated correctly on all ingress pods after the redeploy, but the running HAProxy backend state is not updated on all pods. Some ingress replicas keep stale backend pod IPs in the live runtime state, mark them L4TOUT, and return 503. Other ingress replicas have the correct backend list and work normally.

As a workaround, adding:

`haproxy.org/standalone-backend: "true"`

to the Ingress makes the live backend update correctly after redeploy, and the problem disappears in our testing.

**Environment**

- `haproxytech/kubernetes-ingress chart version: 1.49.0`
- AKS 1.34.2 
- Argo CD

HAProxy Kubernetes Ingress values
```
destNamespace: haproxy-ingress
appName: haproxy-ingress
haproxy-ingress:
  controller:
    ingressClassResource:
      name: haproxy-ingress
    ingressClass: haproxy-ingress
    replicaCount: 4
    service:
      type: LoadBalancer
      externalTrafficPolicy: Local
      annotations:
        service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path: /healthz
        service.beta.kubernetes.io/azure-load-balancer-resource-group: redacted
        service.beta.kubernetes.io/azure-load-balancer-ipv4: redacted
    serviceMonitor:
      enabled: true
    config:
      cr-defaults: haproxy-ingress/haproxy-defaults
      cr-frontend-http: haproxy-ingress/haproxy-frontend-http
      cr-frontend-https: haproxy-ingress/haproxy-frontend-https
      global-config-snippet: |
        ssl-default-bind-options ssl-min-ver TLSv1.2 ssl-max-ver TLSv1.3 no-tls-tickets
        ssl-default-bind-ciphers ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384
        ssl-default-bind-ciphersuites TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
      syslog-server: "address:stdout, length:8192, format:raw, facility:daemon"
    extraArgs:
[...]

PR #802, opened 2026-04-09
BUG/MAJOR: service: deep-copy cr-backend spec to fix cross-service server contamination

cxfcxf:
https://github.com/haproxytech/kubernetes-ingress/issues/795
we have same bug, this fix the bug

this cause way bigger issue since we run multiple backends with external mode.

the actually backend SRV would duplicate and override each other, its not really a drift but totally broke when pod ip gets changed on a rolling upgrade case

deepcopy would avoid this

the actually block that cause it
https://github.com/haproxytech/kubernetes-ingress/blob/master/pkg/service/service.go#L227-L238

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -200,6 +200,16 @@ func (s *Service) getBackendModel(store store.K8s, a annotations.Annotations, cl
 	// get/create backend Model
 	backend, err = annotations.ModelBackend("cr-backend", s.resource.Namespace, store, s.annotations...)
 	logger.Warning(err)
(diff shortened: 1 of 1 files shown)


mdecalf:
Can we expect it to be merged soon?

oktalz:
I'm sorry for the delays, in about a week this should be handled

mdecalf:
Any news on this one?

ivanmatmati:
Hi @cxfcxf, Thanks for PR. Now merged in the codebase.

Issue #807, opened 2026-04-21
Basic Auth - 401 Unauthorized


zurrehma:
Hi,
I am trying to enable basic auth for my HAproxy ingress resource but each time i am getting below error.

```
curl -u myuser:mypassword https://example.com/ -k
<html><body><h1>401 Unauthorized</h1>
You need a valid user and password to access this content.
</body></html>
```

```
kubectl create secret generic haproxy-credentials `
  --from-literal=myuser=$1$yFwoa5bC$S69wgO19hcVEWUNMsgbRj. `
  -n observability
```

My Config:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: my-ingress
  namespace: my-namespace
  annotations:
    # kubernetes.io/ingress.class: haproxy

    # 🔐 Basic Auth
    haproxy.org/auth-type: basic-auth
    haproxy.org/auth-secret: my-namespace/haproxy-credentials 
    haproxy.org/auth-realm: "Protected OTEL"
    acme.cert-manager.io/http01-edit-in-place: "true"
    cert-manager.io/cluster-issuer: letsencrypt-prod
    ssl-redirect: "true"

spec:
  ingressClassName: haproxy
  tls:
  - hosts:
    - example.com
    secretName: my-tls
  rules:
  - host: example.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: my-svc
            port:
              number: 4310
```

PR #821, opened 2026-06-23
BUG/MINOR: rules: apply SSL redirect before basic auth

dragoangel:
## Summary

Fixes ordering of HTTP request rules so redirect executes before auth for ingress flows using SSL redirect.

## Problem

Current ordering can challenge basic auth on HTTP first, then redirect, which is insecure and causes wrong behavior for issue #807 scenario.

## What changed

- Reordered rule priority to place redirect before auth in `types.go`
- Added rule-priority regression test in `types_test.go`
- Added end-to-end rules-assembly regression test in `order_refresh_test.go` validating final order is: `redirect`, then `auth`

## Validation
`go test ./pkg/haproxy/rules`

Verified runtime test output order during debugging: redirect first, auth second.

## Impact

Behavior change is ordering of redirect vs auth vs other directives.

## Other options considered

A possible alternative to rule reordering is to avoid rendering basic auth on the HTTP (non-TLS) frontend when SSL redirect is enabled for the ingress. In that model, the HTTP frontend would only perform redirect and would not pass traffic directly, so basic auth on HTTP would not be needed. If an ingress has basic auth but does not enable TLS redirect, then basic auth would still be rendered on HTTP for that ingress.

## Trade-offs

The current PR approach (redirect before auth) also makes redirect happen before other HTTP request rules such as rate limits. This may be desirable for security/UX, but it can also change expected behavior for users who rely on rate limiting before redirect. For transparency, this alternative is highlighted as a valid implementation path and, if maintainers prefer it, I can implement it either in this PR or in a dedicated follow-up PR.

## Expected result 

Users are redirected to HTTPS before basic-auth challenge.


dragoangel:
Hi @oktalz, can you please help with review?

ivanmatmati:
Hi @dragoangel , LGTM. Just fix the issue with commit message, thanks.

dragoangel:
Thanks @ivanmatmati - will check this and failed e2e

dragoangel:
@ivanmatmati can you please approve workflow run? commit fixed

oktalz:
seems e2e is not working properly on github, I'll try it locally

oktalz:
all good, I need to do some additional checks, but e2e failing is unrelated to changes

oktalz:
@dragoangel merged, thx!

dragoangel:
@oktalz thank you to 🔥

Issue #770, opened 2026-01-13
`whitelist`/`allow-list` works intermittently


mateuszdrab:
Hello

After some recent improvements in my monitoring infrastructure, I've confirmed (an old concern that) my ingresses that have been annotated with a whitelist are intermittently available to non-whitelisted IP addresses.

The annotation I have tried using is:

`haproxy.org/allow-list: "192.168.0.0/16, 10.0.0.0/8"` (also tried whitelist)

Haproxy service and backing infrastructure are configured so that originating IP addresses for connections are correctly passed through all the way to the pod, so that each connection is logged with correct source IP. 

Unfortunately, despite this configuration, I am observing intermittency in enforcement of the whitelist - when working correctly, I observe a 403 on requests from disallowed IPs, when not working, I get standard 200 OK with the content of the page. Tried with curl and blackbox monitor, and haproxy logs confirm that at the time the enforcement isn't working, source IPs are outside of the whitelisted range (initially wanted to rule out an internal network misconfiguration).

I have noticed that inside the generated haproxy.cfg file I can see the line (when grepping for 403):

`http-request deny deny_status 403 if { var(txn.path_match) -m dom 5cecb88cc44beded7e19977ae5fdea99 } !{ src -f /etc/haproxy/maps/allowlist-6e90f8b0f4302b42ab5725b46cf65c5a.map }` 

This line, however, sometimes appears once and sometimes twice. When testing, most of the time it disappears from the https frontend generated config, resulting in https requests to my host not respecting the whitelist, whilst http requests do.

The content of the file `/etc/haproxy/maps/allowlist-6e90f8b0f4302b42ab5725b46cf65c5a.map` is 
```
10.0.0.0/8
192.168.0.0/16
```

There is no pattern to this, sometimes the enforcement works for half an hour, then doesn't for 5 minutes, other times, it's the opposite.

This is on the latest version of the ingress controller.

Any ideas as to why this intermittency would happen?

I will try to run an instance of the controller in debug mode to see if I can get more insight on the issue.

Thanks

mateuszdrab:
Hey @hdurand0710 

Apologies, I'd love to be able to answer this, it would the be within the past 2 years. Before then, I was not using the `whitelist` annotation as this particular service was open 'to the wild'.

I remember a couple of occasions where I loaded up the URL from my phone, and it worked, I thought, that's bizarre, but then it would it would start showing 403s, so I thought it was temporary (or caching from when I was at home on LAN).

[...]

PR #831, opened 2026-07-13
failing test on master regarding ssl-passthrough

KlausNie:
## Purpose

**Test-only demonstration PR — intentionally contains no fix, and the new e2e test is expected to FAIL against master.** It reproduces a fail-open bug in allow-list/deny-list enforcement; a full write-up will follow as an issue linking back here.

## What it demonstrates

Once any single ingress in the cluster uses `ssl-passthrough`, the controller-wide `haproxy.SSLPassthrough` flag is `true`, and `addRules()` (`pkg/ingress/ingress.go:167-170`) places every ingress's `REQ_DENY`/`REQ_CAPTURE` rule on `{http, ssl}` instead of `{http, https}` — dropping the `https` frontend even for ingresses that never opted into passthrough. Their TLS traffic still terminates on the `https` frontend (relayed via the `ssl` frontend's default backend, or arriving directly on the QUIC binds), so their `allow-list`/`deny-list` silently goes unenforced on HTTPS while plain HTTP stays correctly filtered.

This is distinct from #770 / 79d754ae: that fix made the frontend selection *stable* per reconcile pass (it previously depended on Go map iteration order mid-pass), but the stable selection itself still drops the `https` frontend. Observed live on 3.2.12: the intermittent enforcement described in #770 became a permanent fail-open.

### DIFF
--- a/deploy/tests/e2e/https/allowlist_passthrough_test.go
+++ b/deploy/tests/e2e/https/allowlist_passthrough_test.go
@@ -1,0 +1,123 @@
+// Copyright 2026 HAProxy Technologies LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


ivanmatmati:
Hi @KlausNie , we have produced a fix for the bug along with an integration test. Do you see any extra value to add yours at this point ?

Issue #832, opened 2026-07-13
REQ_DENY/REQ_CAPTURE rules drop the https frontend for non-passthrough ingresses once any ingress in the cluster uses ssl-passthrough


KlausNie:
### Summary
`haproxy.SSLPassthrough` is a controller-wide (process-global) flag: it becomes true as soon as any single ingress in the cluster sets `ssl-passthrough: true`, not just the ingress currently being processed.
[`addRules()` (`pkg/ingress/ingress.go`)](https://github.com/haproxytech/kubernetes-ingress/blob/fc1206761aa7948e051eebd56c736802e1d7e226/pkg/ingress/ingress.go#L167-L170) uses this flag to decide which frontends an ingress's REQ_DENY/REQ_CAPTURE rule (i.e. its allow-list/deny-list or capture annotation) gets attached to:
```go
case rules.REQ_DENY, rules.REQ_CAPTURE:
    if haproxy.SSLPassthrough {
        frontends = []string{h.FrontHTTP, h.FrontSSL}   // FrontHTTPS dropped
    }
```
When the flag is true, FrontHTTPS is dropped from the frontend list for every ingress carrying such a rule including ingresses that never opted into ssl-passthrough themselves. But a non-passthrough ingress's TLS traffic still terminates on FrontHTTPS: [`HTTPS.toggleSSLPassthrough` (`pkg/handler/https.go`)](https://github.com/haproxytech/kubernetes-ingress/blob/fc1206761aa7948e051eebd56c736802e1d7e226/pkg/handler/https.go#L275) replaces FrontHTTPS's public binds with an internal PROXY-protocol unix socket once SSLPassthrough is active, and FrontSSL's default backend (ssl-backend) forwards any non-passthrough SNI there for termination. So that traffic ends up completely unchecked by its own allow-list/deny-list.

### Environment
- Controller version: 3.2.9 through 3.2.12 (verified at the exact lines in both tags), still present on master@fc120676
- Chart: kubernetes-ingress-1.52.0 / 1.52.1
- Observed on a cluster with two ingresses on different hosts: one plain (haproxy.org/allow-list, no passthrough) and one with ssl-passthrough: true (fronting Pinniped Concierge's impersonation proxy, which needs the raw mTLS handshake to terminate at the backend for client-cert auth passthrough is a hard requirement there, not incidental)

### Impact observed
With the allow-list ingress and the passthrough ingress coexisting:
[...]

PR #836, opened 2026-08-08
BUG/MEDIUM: ssl-passthrough: keep allow/deny-list on the https frontend

KlausNie:
## Problem
This fixes issue #832. `haproxy.SSLPassthrough` is controller-wide: it becomes true when any ingress uses `ssl-passthrough`. While it is true, `addRules()` dropped the `https` frontend from every ingress's allow-list/deny-list rule. But a non-passthrough ingress still terminates its TLS on that frontend (via `FrontSSL`'s default backend). So the allow-list failed open on HTTPS and QUIC, while plain HTTP stayed filtered. This differs from `79d754ae`, which only made the flag value stable in one pass.

## Fix
Keep `FrontHTTPS` in the frontend list. For a non-passthrough ingress the extra `FrontSSL` placement does nothing, because its host is never in the SNI map.

## Test
An `ssl-passthrough` ingress runs next to a plain allow-list ingress. The test asserts the allow-list ingress returns 403 over HTTPS for an off-list client, and 200 after the allow-list widens to `0.0.0.0/0`.

Supersedes #831 (test-only demonstration).

### DIFF
--- a/deploy/tests/e2e/https/allowlist_passthrough_test.go
+++ b/deploy/tests/e2e/https/allowlist_passthrough_test.go
@@ -1,0 +1,123 @@
+// Copyright 2026 HAProxy Technologies LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


KlausNie:
@imatmati the complete PR for https://github.com/haproxytech/kubernetes-ingress/issues/832

let me know, if you'd like something changed!

oktalz:
@KlausNie thx!

PR #838, opened 2026-08-14
BUG/MINOR: e2e: build host-port addresses with net.JoinHostPort

MaxRink:
`go vet ./...` reports on unmodified master:

```
deploy/tests/e2e/client.go:200:43: address format "%s:%d" does not work with IPv6
```

CI does not run `go vet`, so the defect survived.

## Changes

**`deploy/tests/e2e/client.go`** builds three addresses from `KIND_URL` with `fmt.Sprintf("%s:%d", ...)`, at lines 84, 156 and 200. An IPv6 literal in `KIND_URL` yields an unparsable address, so `DialContext`, `ResolveTCPAddr` and `Dial` fail. `net.JoinHostPort` brackets the literal. Only line 200 is visible to `go vet`, the other two are the same defect.

**`pkg/gateways/gateways.go`** builds the `serversByBackend` change-detection key with the same format. With IPv6 the key is ambiguous, so a real endpoint change can escape the reload comparison. Same fix.

**`taskfile.yml`** adds `go vet ./...` to the `lint` task, which the `linters` CI job runs. `go vet` is clean across the tree after this change, with and without the e2e build tags, so it adds no noise.

Not changed: `main.go:229` and `:232` format `IPV4BindAddr`, which is IPv4 by definition and only logged. `pkg/utils/logging.go` and `pkg/haproxy/instance/configuration.go` format `file:line`, not host-port.

## Verification

`go vet ./...`, `go build ./...`, revive v1.13.0 with `revive.toml`, staticcheck v0.6.1 and gofumpt are all clean. `go test ./pkg/... ./test/...` passes. `go vet` and `go build` with the e2e build tags are clean.


oktalz:
thx @MaxRink !

PR #849, opened 2026-09-01
CLEANUP/MINOR: annotations: skip empty elements in allow-list and deny-list

### LINKED ISSUE #848: Trailing comma in haproxy.org/allow-list results in access being allowed from anywhere
## Description
A trailing comma in the `haproxy.org/allow-list` annotation appears to cause the allow-list to behave as if it were empty, effectively allowing traffic from **any source IP**.

For example:

```yaml
metadata:
  annotations:
    haproxy.org/allow-list: "10.0.0.0/8,192.168.1.0/24,"
```

Without the trailing comma, access is correctly restricted to the specified networks:

```yaml
haproxy.org/allow-list: "10.0.0.0/8,192.168.1.0/24"
```

However, adding the trailing comma causes the Ingress to become accessible from arbitrary external IP addresses.

## Expected behavior

A trailing comma should either:

* be ignored, resulting in the same configuration as without the trailing comma; or
* cause the annotation to be rejected as invalid.

It should **not** result in the allow-list being disabled or interpreted as allowing all sources.

## Actual behavior

With:

```yaml
haproxy.org/allow-list: "10.0.0.0/8,192.168.1.0/24,"
```

the configured source restrictions are effectively bypassed and the service becomes reachable from the public internet.

This is particularly concerning because a trailing comma can easily be introduced accidentally when generating the annotation dynamically.

## Security impact
This can result in an unintended exposure of an otherwise restricted Kubernetes Ingress.

For example, a configuration intended to allow only internal networks:

```yaml
haproxy.org/allow-list: "10.0.0.0/8,172.16.0.0/12,"
```

can potentially become publicly accessible simply because of the trailing comma.

This could lead to accidental exposure of internal services.

## Reproduction
Create an Ingress with:

```yaml
kind: Ingress
metadata:
  name: test
  annotations:
    haproxy.org/allow-list: "10.0.0.0/8,"
spec:
  ingressClassName: haproxy
  # ...
```

Then access the Ingress from an IP outside the allowed `10.0.0.0/8` network.

The request is allowed, whereas removing the trailing comma correctly blocks the request.

## Environment

* HAProxy Kubernetes Ingress Controller: Community Edition
* Controller version: `3.2.6`
* HAProxy version: `3.2.12`

Given the security implications, it may also be worth treating malformed `allow-list` annotations as configuration errors and failing closed rather than falling back to allowing all traffic.

Reference:

[HAProxy Kubernetes Ingress Controller — allow-list documentation](https://www.haproxy.com/documentation/kubernetes-ingress/community/configuration-reference/ingress#allow-list)

locker95:
A trailing comma in the allow-list annotation, e.g. `haproxy.org/allow-list: "10.0.0.0/16,"`, makes `strings.Split` produce an empty element. `net.ParseCIDR("")` then fails and the whole annotation is discarded with `incorrect address '' in allow-list annotation`, so the ingress silently ends up open to all traffic instead of being restricted. The same applies to deny-list and the deprecated whitelist/blacklist names.

This patch skips elements that are empty after trimming, so trailing commas, consecutive commas or stray whitespace no longer invalidate the list. A value with no valid element at all is still rejected as before: adding the rule with an empty map would flip an allow-list into denying everyone, which is just as surprising as allowing everyone.

Added a table-driven test for AccessControl.Process covering trailing/consecutive commas, whitespace-only elements, the deprecated whitelist name, deny-list, pattern files and the existing error paths.

Fixes #848


pkg/annotations/ingress/accessControl.go
```diff
@@ -63,14 +63,24 @@ func (a *AccessControl) Process(k store.K8s, annotations ...map[string]string) (
...
 
 	if !a.maps.MapExists(mapName) {
+		hasValidAddress := false
 		for _, address := range strings.Split(input, ",") {
 			address = strings.TrimSpace(address)
+			if address == "" {
+				// skip empty elements coming from a trailing comma,
+				// consecutive commas or surrounding whitespace
+				continue
+			}
```
oktalz, line 71: it better to just use strings.FieldsFunc(input, ",") in loop instead of adding more complexity, hasValidAddress is still OK

locker95:
switched the split to FieldsFunc so trailing/double commas drop out. still trim spaces so `1.2.3.4, 5.6.7.8` works.

oktalz:
as mentioned in https://github.com/haproxytech/kubernetes-ingress/issues/848 I needed to change overall commit message I hope that change is fine by you ?

I pushed the change already on a PR

@locker95 is that OK for you?

locker95:
yeah that's fine, thanks for fixing it up.

PR #856, opened 2026-09-21
Add dynamic namespace selection by label

### LINKED ISSUE #853: Proposal: dynamic namespace selection by label
Hello maintainers,

Would you be open to reconsidering support for selecting watched namespaces with a Kubernetes label selector? This follows #847 and the earlier implementation in #852, which are now closed. I have since reworked the design and expanded the test coverage.

### Use case

The existing `--namespace-whitelist` and `--namespace-blacklist` options are based on namespace names fixed at startup. In our clusters, namespace membership follows labels and changes over time. Keeping a generated whitelist in sync requires changing controller arguments and restarting the controller.

### Proposed behavior

A flag such as:

```bash
--namespace-label-selector='team=backend'
```

would provide this behavior:

- Matching namespaces are watched at startup.
- Adding a matching label starts watching that namespace without restarting the controller.
- Removing or changing the label stops its namespaced watchers and removes its resources from the generated HAProxy configuration.
- Relabeling the namespace starts a fresh list/watch, so changes made while it was unselected are picked up.
- If a whitelist or blacklist is configured, the existing behavior wins and the selector is ignored with a warning.

### Implementation

The revised implementation keeps one watch session per selected namespace. Each session owns its informer factories and has a monotonically increasing epoch. Events from an old session are rejected after an unlabel/relabel cycle. A namespace becomes eligible for configuration only after its initial informer events have been processed.

Namespace discovery still uses a server-side label selector. The default, whitelist, and blacklist paths keep their existing behavior.

### Validation

The current branch includes:

- Unit coverage for session lifecycle, stale epochs, teardown, bootstrap ordering, late CRD registration, TCP services, Gateway references, and non-selector compatibility.
- Race tests for the session, store, and handler packages.
- Sequential Kubernetes e2e coverage for startup selection, label/unlabel/relabel, two matching namespaces, bootstrap and rapid churn, namespace deletion, TCP CRs, late CRD discovery, and whitelist/blacklist precedence.

All of those checks currently pass.

### Trade-offs

This creates informer factories per selected namespace, so API traffic and goroutine count grow with the number of matches. The intended use is a controller scoped to one tenant or a bounded set of namespaces, not a selector that matches an entire large cluster. Label changes become visible on the next normal sync cycle, up to `--sync-period` (5 seconds by default).

Would this direction fit the project? If so, I can rebase the implementation on the current master branch and open a focused PR for review. I would also appreciate guidance on any behavior or test expectations you would like changed before that.

Thank you for your time.

fourhu:
Fixes #853.

## Summary

Add `--namespace-label-selector` for clusters where namespace membership is controlled by labels instead of a fixed name list.

Matching namespaces are watched at startup. Labeling a namespace starts its watches without restarting the controller; removing or changing the label removes its resources from HAProxy configuration. Relabeling starts a fresh list/watch so changes made while the namespace was unselected are loaded.

`--namespace-whitelist` and `--namespace-blacklist` keep precedence. If either is configured, the selector is ignored with a warning. With the selector unset, the existing informer path is unchanged.

## Implementation

- Keep one informer session per selected namespace.
- Assign each session a monotonically increasing epoch and reject events from retired sessions.
- Admit a namespace to HAProxy configuration only after its initial informer events have been processed.
- Handle label churn, namespace deletion and recreation, late CRD registration, TCP resources, and Gateway API references.
- Document the flag and precedence rules.

## Validation

The branch adds unit and sequential Kubernetes e2e coverage for startup selection, label and unlabel transitions, relabeling, stale sessions, namespace deletion and recreation, TCP CRs, late CRDs, and whitelist/blacklist precedence.

## Trade-off

The selector path creates informer factories per selected namespace. API traffic and goroutine count therefore grow with the number of matches. This option is intended for a bounded namespace set; leaving the flag unset keeps the current shared-informer behavior.


fourhu:
Superseded by a new PR with an accurate description of the current selector behavior.

